PR #4, opened 2022-09-19
Build distroless container images

### LINKED ISSUE #1: [BUG] Remove dependence on `lspci` to enable distroless containers
# Problem
Currently, pcidevices controller is built on an Alpine Linux container. It is not possible to make it a distroless container the way it currently works.

# Solution
Replace using `os/exec` on the `lspci` command with the `jaypipes/pci` library.

### LINKED ISSUE #2: [BUG] PCIDeviceStatus.KernelModules is not being populated
# Problem
The [KernelModules](https://github.com/harvester/pcidevices/blob/master/pkg/apis/devices.harvesterhci.io/v1beta1/pcidevice.go#L35) field isn't being populated. Possibly a problem with libkmod

# Solution
Either remove the field or figure out why the kernel module list isn't being populated.

eliaskoromilas:
fixes #1, fixes #2

**- What I did**

* Drop external binary dependencies
* Build distroless container images

**- How I did it**

* Using `jaypipes/ghw/pkg/pci` instead of executing `lspci`
* Using `u-root/u-root/pkg/kmodule` instead of executing `lsmod/modprobe`
* Embedding https://pci-ids.ucw.cz/v2.2/pci.ids in the distroless container image

**- How to verify it**

Unit tests are updated.


eliaskoromilas:
**Notes**:

I didn't change any of the logic, just replaced the libraries. The auto-generated code has some updates -`jaypipes` uses strings for the PCI device identifiers and I adopted that cause it looked more aligned with the rest of the code.

I couldn't find a way (library) to scrape `KernelModules` in Go, but since it doesn't work inside containers, I would suggest removing it here completely.

tlehman (commented):
I tested these changes, new container images work

PR #11, opened 2022-09-21
drop requirement for binary release

ibrokethecloud:
Minor change in drone ci to drop step for releasing binary.


guangbochen (approved):
LGTM

PR #12, opened 2022-09-21
Merge and revisit dev manifests

eliaskoromilas:
Apart from making it easier for devs to test this, the main change in this PR is the addition of the `/sys` host path volume mount. I couldn't make `pcideviceclaims` work in my system without it (the controller couldn't write under `/sys/bus/pci/drivers`). Tested on Harvester v1.0.3 with `iommu` enabled.

```
time="2022-09-21T16:55:23Z" level=info msg="Attempting to enable passthrough"
time="2022-09-21T16:55:23Z" level=error msg="PCI Device Claim reconciliation error: open /sys/bus/pci/drivers/vfio-pci/new_id: read-only file system"

### CODE REVIEW

File: manifests/pcidevices-operator.yaml [hunk 1]
```diff
@@ -0,0 +1,124 @@
...
+              memory: 200Mi
+            requests:
+              cpu: 10m
+              memory: 50Mi
+          securityContext:
+            privileged: true
+          volumeMounts:
+            - mountPath: /lib/modules
+              name: modules
+            - mountPath: /sys
```
tlehman, line 88: I'm curious how this worked before without the /sys mount

tlehman (approved):
LGTM

manifests/pcidevices-operator.yaml
eliaskoromilas: It might sound crazy, but it was working thanks to `hostNetwork: true`!

PR #13, opened 2022-09-21
Fix permissions and name errors FIXES #2802

tlehman:
I tested this on my 2 node cluster after installing the 1.1.0-rc1 (version 008b5d5)

This works, I verified that the PCI Devices create, that PCI Device Claims prepare the devices for passthrough.

Also I fixed a regression introduced in this PR: https://github.com/harvester/pcidevices/pull/12


ibrokethecloud (approved):
LGTM

PR #14, opened 2022-09-22
After reboot, rebind pdc devices to vfio-pci FIXES #2819

tlehman:
Fix #2819 by looking at all PCIDeviceClaims whose underlying devices are not currently bound to vfio-pci. This usually happens after a node reboot.

# Other small changes
- hostname switched to more accurate "nodename"

# Testing
I created 6 PCIDeviceClaims, all 6 underlying devices were correctly bound to `vfio-pci`, and then rebooted. Those six devices were not bound to `vfio-pci` anymore. I added logic into the reconciliation loop that unbound the PCI Devices belonging to the claims, and then bound them to `vfio-pci`. I verified that it worked on my 6 underlying devices.

### DIFF
--- a/pkg/controller/pcideviceclaim/pcideviceclaim_controller.go
+++ b/pkg/controller/pcideviceclaim/pcideviceclaim_controller.go
@@ -205,10 +205,6 @@ func (h Handler) reconcilePCIDeviceClaims(nodename string) error {
 					}
 					pdc.Status.PassthroughEnabled = true
 				}
-				_, err = h.pdcClient.Update(&pdc)
-				if err != nil {
-					return err
-				}
 				_, err = h.pdcClient.UpdateStatus(&pdc)
 				if err != nil {
 					return err
--- a/pkg/controller/pcideviceclaim/pcideviceclaim_controller.go
+++ b/pkg/controller/pcideviceclaim/pcideviceclaim_controller.go
@@ -29,12 +29,12 @@ type Handler struct {
 func Register(
 	ctx context.Context,
 	pdcClient v1beta1gen.PCIDeviceClaimController,
-	pd v1beta1gen.PCIDeviceController,
+	pdClient v1beta1gen.PCIDeviceController,
 ) error {
 	logrus.Info("Registering PCI Device Claims controller")
 	handler := &Handler{
 		pdcClient: pdcClient,
-		pdClient:  pd,
+		pdClient:  pdClient,
 	}
 	nodename := os.Getenv("NODE_NAME")
 	// start goroutine to regularly reconcile the PCI Device Claims' status with their spec
@@ -154,6 +154,8 @@ func (h Handler) reconcilePCIDeviceClaims(nodename string) error {
 					err = addNewIdToVfioPCIDriver(pd.Status.VendorId, pd.Status.DeviceId)
 					if err != nil {
 						logrus.Errorf("Error rebinding device after reboot: %s", err)
+					} else {
+						pdc.Status.PassthroughEnabled = true
 					}
 				}
 			}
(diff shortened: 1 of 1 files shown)


pkg/controller/pcideviceclaim/pcideviceclaim_controller.go
```diff
@@ -133,20 +133,28 @@ func (h Handler) reconcilePCIDeviceClaims(hostname string) error {
...
 		// After reboot, the PCIDeviceClaim will be there but the PCIDevice won't be bound to vfio-pci
-		if found && pd.Status.KernelDriverInUse != "vfio-pci" && hostname == pd.Status.NodeName {
+		if pd.Status.KernelDriverInUse != "vfio-pci" && nodename == pd.Status.NodeName {
 			// Set PassthroughEnabled to false
 			for _, pdc := range pdcs.Items {
 				if pdc.Spec.Address == pd.Status.Address {
 					logrus.Infof("Passthrough disabled for device %s", pd.Name)
 					pdc.Status.PassthroughEnabled = false
+					err = unbindPCIDeviceFromDriver(pd.Status.Address, pd.Status.KernelDriverInUse)
+					if err != nil {
```
w13915984028, line 151: When `err != nil`, should return or not ?

if `unbindPCIDeviceFromDriver` failed but not return, then what will happen in `addNewIdToVfioPCIDriver` ?

pkg/controller/pcideviceclaim/pcideviceclaim_controller.go
```diff
@@ -133,20 +133,28 @@ func (h Handler) reconcilePCIDeviceClaims(hostname string) error {
...
+						logrus.Errorf("Error unbinding device after reboot: %s", err)
+					}
+					err = addNewIdToVfioPCIDriver(pd.Status.VendorId, pd.Status.DeviceId)
+					if err != nil {
+						logrus.Errorf("Error rebinding device after reboot: %s", err)
+					}
```
eliaskoromilas, line 157: should we set `PassthroughEnabled = true` here?

w13915984028 (commented):
More questions are asked in other places.

pkg/controller/pcideviceclaim/pcideviceclaim_controller.go
eliaskoromilas: it will fail, we could `continue` here
sometimes there is no driver to unbind, as is, the function will fail, we should skip `unbindPCIDeviceFromDriver` in those cases

pkg/controller/pcideviceclaim/pcideviceclaim_controller.go
tlehman: Yes, good catch

pkg/controller/pcideviceclaim/pcideviceclaim_controller.go
tlehman: That's right, I have worked with devices that didn't have a driver to unbind, we want to continue executing even if the unbind function fails

w13915984028: With those exceptions, then it looks hard to identify the `devices that didn't have a driver to unbind`  and  `unbind failed`.

It is better to add comment here.  And the `logrus.Errorf ` needs also say like ` ..., the error is skipped`, maybe `logrus.Warnf` is the proper level.

Without those, pepole will have question to the code, the POD LOG will also bring surprise.

w13915984028:
Question:

Should `if pdc.DeletionTimestamp != nil` be put before `if !pdc.Status.PassthroughEnabled` ?  It's normal to check the timestamp at first.  When it's to be deleted, the `enable` seems make no sense.

```
reconciler:

			if !pdc.Status.PassthroughEnabled {
				logrus.Infof("Attempting to enable passthrough")
...
			}

			if pdc.DeletionTimestamp != nil {
				logrus.Infof("Attempting to unbind PCI device %s from vfio-pci", pdc.Spec.Address)
                        }
```

PR #20, opened 2022-10-05
cleanup reconcile logic

ibrokethecloud:
Minor changes to fix up error handling and deletion of objects.

### DIFF
--- a/pkg/controller/pcidevice/pcidevice_controller.go
+++ b/pkg/controller/pcidevice/pcidevice_controller.go
@@ -9,7 +9,9 @@ import (
 	ctl "github.com/harvester/pcidevices/pkg/generated/controllers/devices.harvesterhci.io/v1beta1"
 	"github.com/jaypipes/ghw"
 	"github.com/sirupsen/logrus"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/labels"
 )
 
 const (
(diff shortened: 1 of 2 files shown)


pkg/controller/pcidevice/pcidevice_controller.go
```diff
@@ -49,42 +51,68 @@ func (h Handler) reconcilePCIDevices(nodename string) error {
...
+	for _, v := range pdList.Items {
+		if ok := setOfRealPCIAddrs[v.Status.Address]; !ok {
+			deleteList = append(deleteList, v)
+		}
+	}
+
+	for _, v := range deleteList {
+		if err := h.client.Delete(v.Name, &metav1.DeleteOptions{}); err != nil {
+			logrus.Errorf("[PCIDeviceController] Faield to delete non existent device: %s on node %s", v.Name, v.Status.NodeName)
+			return err
```
Vicente-Cheng, line 113: Is it too aggressive that we need to break down here when any delete operation fails?

This section does the work named `// remove non-existent devices`.
If we ignore the delete operation fails, that we would get the non-existent device on the client?
Just curious, What problems would cause by non-existent devices?

pkg/controller/pcidevice/pcidevice_controller.go
```diff
@@ -49,42 +51,68 @@ func (h Handler) reconcilePCIDevices(nodename string) error {
...
-
-			// Create the PCIDevice CR if it doesn't exist
-			var pdToCreate v1beta1.PCIDevice = v1beta1.NewPCIDeviceForHostname(dev, nodename)
-			logrus.Infof("Creating PCI Device: %s\n", err)
-			pdToCreate.Labels = map[string]string{"nodename": nodename} // label
-			_, err := h.client.Create(&pdToCreate)
-			if err != nil {
-				logrus.Errorf("Failed to create PCI Device: %s\n", err)
+			if apierrors.IsNotFound(err) {
+				logrus.Infof("[PCIDeviceController] Device does %s does not exist: %s\n", name)
```
bk201, line 65:
```suggestion
				logrus.Infof("[PCIDeviceController] Device %s does not exist: %s", name)
```

We don't need `\n` in the logging messages.

pkg/controller/pcidevice/pcidevice_controller.go
ibrokethecloud: I would assume the object exists and is available in the harvester UI for pcidevice claims and passthrough when in reality the backing physical device is now missing.

Vicente-Cheng: OK, that makes sense to me. Thanks for explaining.

pkg/controller/pcidevice/pcidevice_controller.go
Vicente-Cheng: Still have some redundant `\n` need to remove.

https://github.com/harvester/pcidevices/pull/20/files#diff-8c6dfb3b3b6b632d555b11e4d7d51828f55bfacd790ef5a42edd955a8437f692R73
https://github.com/harvester/pcidevices/pull/20/files#diff-8c6dfb3b3b6b632d555b11e4d7d51828f55bfacd790ef5a42edd955a8437f692R89

pkg/controller/pcidevice/pcidevice_controller.go
```diff
@@ -49,42 +51,68 @@ func (h Handler) reconcilePCIDevices(nodename string) error {
...
+
+				// Create the PCIDevice CR if it doesn't exist
+				var pdToCreate v1beta1.PCIDevice = v1beta1.NewPCIDeviceForHostname(dev, nodename)
+				logrus.Infof("Creating PCI Device: %s\n", err)
+				pdToCreate.Labels = commonLabels
+				_, err := h.client.Create(&pdToCreate)
+				if err != nil {
+					logrus.Errorf("[PCIDeviceController] Failed to create PCI Device: %v\n", err)
+					return err
+				}
```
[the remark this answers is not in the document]
ibrokethecloud: if the object cant be created.. then next reconcile pass will pick it up again and try again.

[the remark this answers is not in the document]
Vicente-Cheng: So, it means we would skip this round when PCI device CR was just created?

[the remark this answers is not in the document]
Vicente-Cheng: OK... That's polling for reconcile.
It seems to work with next round.

Vicente-Cheng (approved):
overall LGTM. And please squash some redundant commits, thanks!

pkg/controller/pcidevice/pcidevice_controller.go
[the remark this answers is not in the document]
ibrokethecloud: yep.. i have fixed it too. we should be able to reconcile in the first iteration.

Vicente-Cheng (approved):
LGTM with the latest change. Thanks!!

pkg/controller/pcidevice/pcidevice_controller.go
```diff
@@ -49,42 +51,70 @@ func (h Handler) reconcilePCIDevices(nodename string) error {
...
-
-			// Create the PCIDevice CR if it doesn't exist
-			var pdToCreate v1beta1.PCIDevice = v1beta1.NewPCIDeviceForHostname(dev, nodename)
-			logrus.Infof("Creating PCI Device: %s\n", err)
-			pdToCreate.Labels = map[string]string{"nodename": nodename} // label
-			_, err := h.client.Create(&pdToCreate)
-			if err != nil {
-				logrus.Errorf("Failed to create PCI Device: %s\n", err)
+			if apierrors.IsNotFound(err) {
+				logrus.Infof("[PCIDeviceController] Device %s does not exist: %s", name)
```
bk201, line 67: need to arguments here:

```
level=info msg="[PCIDeviceController] Device node3-000000030 does not exist: %!s(MISSING)"
```

ibrokethecloud: good spot. fixed it up.

bk201 (approved):
LGTM! Thanks for addressing comments.

PR #23, opened 2022-10-11
minor changes for todos and more fixes

ibrokethecloud:
This PR builds on top of fixes introduced under: https://github.com/harvester/pcidevices/pull/21 and adds a few minor fixes and wiring changes for the register function.

PR #24, opened 2022-10-11
Retry rebind after reboot if error happens

tlehman:
- Based on feedback from PR #23
- Returning the error from rebindAfterReboot will cause the call to Register to return an error, which will crash the Pod. This is desired behavior, because we want to retry the rebindAfterReboot before doing anything else.


guangbochen (approved):
LGTM, thanks.

PR #25, opened 2022-10-13
change to mutation logic

ibrokethecloud:
PR changes the mutation logic to use Kubevirt VM, PCIDevice info to identify if the pod needs SYS_RESOURCE cap to be added.

Helm Chart needs an update as Cluster role now needs access to `VirtualMachines` in apiGroup `kubevirt.io` with "get,list,watch" permissions for the indexer to work.


tlehman (approved):
LGTM

eliaskoromilas:
Sorry for the late review. I would strongly suggest checking VMIs instead of VMs. I also think that simply calling [`IsVFIOVMI`](https://github.com/kubevirt/kubevirt/blob/v0.54.0/pkg/util/util.go#L72-L79) should be adequate... is cross checking the device cache adding any value?

PR #26, opened 2022-10-17
Change base image to SLE

tlehman:
Issue: https://github.com/harvester/harvester/issues/2954

Change to the SLE base image.

### DIFF
--- a/Dockerfile.dapper
+++ b/Dockerfile.dapper
@@ -1,5 +1,4 @@
 FROM registry.suse.com/bci/bci-base:15.3
-RUN zypper -n install pciutils
 
 ARG DAPPER_HOST_ARCH
 ENV ARCH $DAPPER_HOST_ARCH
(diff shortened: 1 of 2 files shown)


Dockerfile.dapper
```diff
@@ -1,9 +1,10 @@
-FROM golang:1.18.5-alpine3.16
+FROM registry.suse.com/bci/bci-base:15.3
```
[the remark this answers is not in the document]
tlehman: I know but I might as well use the same image. On a box with a clean /var directory, that's one less layer it has to pull from the registry.

package/Dockerfile
```diff
@@ -1,5 +1,4 @@
-FROM gcr.io/distroless/static
+FROM registry.suse.com/bci/bci-base:15.3
+RUN zypper -n install pciutils
 COPY bin/pcidevices /bin/pcidevices
 ENTRYPOINT ["pcidevices"]
-# https://pci-ids.ucw.cz/v2.2/pci.ids mirror
-ADD https://github.com/pciutils/pciids/raw/master/pci.ids /usr/share/hwdata/pci.ids
```
eliaskoromilas, line 5: If I may suggest updating the PCI ID list (`update-pciids`).

tlehman: Good call, pushing up a new revision in a moment

Issue #19, opened 2022-09-28
[Bug] KubeVirt Allow-list doesn't always get refreshed


tlehman:
Changing the kubevirt list of permitted devices will trigger a change to the `.status.allocatable` value. If you add multiple identical devices, it only counts the first one. Also, if you remove the device, the allocatable list is not updated.

**Proposed solution**: remove the device and then re-add it, to trigger an update.

eliaskoromilas:
My proposed solution is to use [`PCIDevicePlugin`](https://github.com/kubevirt/kubevirt/blob/v0.54.0/pkg/virt-handler/device-manager/pci_device.go#L72) and run our own KubeVirt device plugins, instead of messing with KubeVirt's config.

The lifecycle is similar (if not simpler) and it's fully aligned with KubeVirt's practices. Under the hood the device plugin injects an environment variable, into the `virt-launcher` pod, that specifies the PCI bus addresses for pass-through.

This will also allow us to control exactly which PCI devices are getting advertised (only the claimed ones) and which are not.

tlehman:
Now that 1.1.0 is out, I am going to implement this.

PR #27, opened 2022-10-20
Call update on KubeVirt to trigger allow-list update

tlehman:
Fixes:
- [2991](https://github.com/harvester/harvester/issues/2991) Can't remove or reassign a PCI device
- [2820](https://github.com/harvester/harvester/issues/2820) Making a PCI Device Claim doesn't allow-list the device in KubeVirt

How I tested this:
- I built `tlehman/pcidevices:dev`, built a new chart and deployed it to my cluster. I confirmed that the KubeVirt allow list stayed up to date when I enabled and disabled my two GPUs. Also I attached a GeForce 1660 to ubuntu-gpu1, installed the driver, then I removed the 1660 and attached it to suse-gpu1 and booted it up. The bug 2991 is caused by the same problem in 2820

### DIFF
--- a/manifests/pcidevices-operator.yaml
+++ b/manifests/pcidevices-operator.yaml
@@ -31,6 +31,9 @@ rules:
   - apiGroups: ["kubevirt.io"]
     resources: ["kubevirts"]
     verbs: [ "get", "update" ]
+  - apiGroups: ["kubevirt.io"]
+    resources: ["virtualmachines"]
+    verbs: [ "get", "list", "watch" ]
 ---
 apiVersion: rbac.authorization.k8s.io/v1
 kind: ClusterRoleBinding
--- a/pkg/controller/pcideviceclaim/pcideviceclaim_controller.go
+++ b/pkg/controller/pcideviceclaim/pcideviceclaim_controller.go
@@ -233,11 +233,12 @@ func (h Handler) addHostDeviceToKubeVirt(pd *v1beta1.PCIDevice) error {
 			ExternalResourceProvider: false,
 		}
 		kvCopy.Spec.Configuration.PermittedHostDevices.PciHostDevices = append(permittedPCIDevices, devToPermit)
-		_, err := h.virtClient.KubeVirt(DefaultNS).Update(kvCopy)
-		if err != nil {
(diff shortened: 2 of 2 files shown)


ibrokethecloud:
I tried this image however it has made no difference in update to the node.status.allocatable devices list.

```
▶ k get pcideviceclaim
k getNAME               ADDRESS        NODE NAME   USER NAME   KERNEL DRIVER ΤΟ UNBIND   PASSTHROUGH ENABLED
static-000000000   0000:00:00.0   static      admin                                 true
static-000000143   0000:00:14.3   static      admin       igb                       true
(⎈ |local:harvester-system)
~
▶ k get node -o yaml | yq '.items[].status.allocatable'
cpu: "8"
devices.kubevirt.io/kvm: 1k
devices.kubevirt.io/sev: "0"
devices.kubevirt.io/tun: 1k
devices.kubevirt.io/vhost-net: 1k
ephemeral-storage: "79757918556"
hugepages-2Mi: "0"
memory: 32845564Ki
pods: "110"
```

tlehman:
@ibrokethecloud I have been digging into the virt-handler code, and this method, [updatePermittedHostDevicePlugins](https://github.com/kubevirt/kubevirt/blob/v0.54.0/pkg/virt-handler/device-manager/device_controller.go#L161) looks promising. This is where the Kubevirt CR (the `virtConfig`) is read. I just need to find if this causes the node status to get updated. If so, I need to find out why this function is not being called.

tlehman:
The `refreshPermittedDevices` function calls the `updatePermittedHostDevicePlugins` function: https://github.com/kubevirt/kubevirt/blob/a330cd4337ed9f5a7ab8fcbbf2dbaaefc0f82659/pkg/virt-handler/device-manager/device_controller.go#L264

tlehman:
I'm going to work on https://github.com/harvester/pcidevices/issues/19 and try out @eliaskoromilas ' idea of making our own DevicePlugin, this would also give us more control over which PCI devices are being advertised.

tlehman:
Closing this in favor of the [deviceplugins PR](https://github.com/harvester/pcidevices/pull/35)

PR #34, opened 2023-01-09
skip PCI devices for NICs involved in the default harvester bonded interfaces

ibrokethecloud:
PR introduces the following:
* Fake pcidevices client to allow testing of reconcile using ghw-snapshots from an existing node
* A sample snapshot
* Ability to skip pcidevice creation for NIC's which are involved in the mgmt-bo and mgmt-br interfaces in harvester

Related to: https://github.com/harvester/harvester/issues/3199


go.mod
```diff
@@ -17,6 +17,7 @@ require (
 	github.com/stretchr/testify v1.8.0
 	github.com/u-root/u-root v7.0.0+incompatible
 	github.com/urfave/cli/v2 v2.11.1
+	github.com/vishvananda/netlink v1.1.1-0.20211118161826-650dca95af54
```
tlehman, line 20: Can our [existing ghw library](https://github.com/jaypipes/ghw) do this already? If so, let's use that instead of bringing in a redundant library.

pkg/controller/pcidevice/pcidevice_controller_test.go
```diff
@@ -0,0 +1,41 @@
...
+
+const (
+	defaultPCIDeviceSnapshot = "../../../tests/snapshots/linux-amd64-e147d239df014921c6cbb49fbc3d6c41.tar.gz"
+)
+
+func Test_reconcilePCIDevices(t *testing.T) {
+	assert := require.New(t)
+	client := fake.NewSimpleClientset()
+
+	pci, err := ghw.PCI(ghw.WithSnapshot(ghw.SnapshotOptions{
```
tlehman, line 22: Our [existing ghw library](https://github.com/jaypipes/ghw) definitely does this too: https://github.com/jaypipes/ghw/blob/main/pkg/option/option.go#L173-L177

pkg/controller/pcidevice/pcidevice_controller_test.go
```diff
@@ -0,0 +1,41 @@
...
+	}
+
+	err = h.reconcilePCIDevices("TEST_NODE")
+	assert.NoError(err, "expected no error during pcidevice reconcile")
+	// check if GPU device is created. Default Name is TEST_NODE-000008000
+	gpuDevice, err := client.DevicesV1beta1().PCIDevices().Get(context.TODO(), "TEST_NODE-000008000", metav1.GetOptions{})
+	assert.NoError(err, "expected no error while listing GPU")
+	assert.NotNil(gpuDevice, "expected to find a GPU device")
+	_, err = client.DevicesV1beta1().PCIDevices().Get(context.TODO(), "TEST_NODE-000004001", metav1.GetOptions{})
+	assert.True(apierrors.IsNotFound(err), "expected to not find the pci address for 000004001")
```
tlehman, line 40: This test is awesome!

pkg/controller/pcidevice/pcidevice_controller.go
```diff
@@ -117,3 +137,13 @@ func (h Handler) reconcilePCIDevices(nodename string) error {
 
 	return nil
 }
+
+func containsString(elements []string, element string) bool {
```
tlehman, line 141: This seems like something that should be a generic standard library function, I can't find anything like that, but let's keep an eye out for it.

tlehman (changes requested):

go.mod
ibrokethecloud: `jaypipes/ghw` doesnt allow me to query relationship between interfaces and if they are involved in a bond interface.

netlink allows me to query the bo/br interfaces, identity their index and query all other interfaces which may have master index set to index of either bo/br. 

Once the physical interface names are available, `ghw` is used to query their pci addresses and subsequently skip the same from pcidevice creation logic.

pkg/controller/pcidevice/pcidevice_controller_test.go
ibrokethecloud: we are using the same, just have a pre-generated snapshot

tlehman (approved):
LGTM

pkg/controller/pcidevice/pcidevice_controller.go
```diff
@@ -2,6 +2,9 @@ package pcidevice
 
 import (
 	"context"
+	"fmt"
```
bk201, line 5: It would be nice to group imports.

ibrokethecloud: fixed this up

PR #35, opened 2023-01-13
PCIDevicePlugins

tlehman:
# Introducing PCIDevicePlugin

This [deviceplugin](https://kubernetes.io/docs/concepts/extend-kubernetes/compute-storage-net/device-plugins/) solves a number of issues (listed below). It also allows more fine-grained control for the exposure and allocation of particular PCI devices to VMs.

# How to test
Find devices that have more than one device with the same resourceName: 

```
kubectl get pcidevices -o yaml | yq '.items[].status | .resourceName' | sort  | uniq -c
```

Then pick one of the resourceNames with more than 1 on the left of the above command output.

Select one, enable it for passthrough, then look at the node status `allocatable` section to verify that the there is 1. Then enable the second one, and verify that the `allocatable` count goes up to 2. Then disable the devices and verify that the `allocatable` count goes down as expected.

# Notes on implementation

When a PCI Device on a Node is enabled for passthrough, you might notice that all the other PCI Devices that share the same `vendorId:deviceId` pair are loaded into the PCIDevicePlugin at initialization. Initially, the one that is enabled for passthrough is marked as "healthy". Then, as other identical devices on that Node are enabled for passthrough, those devices will be marked as healthy. This makes sense if you consider how kubevirt uses the healthy/unhealthy distinction. Kubevirt's PCIDevicePlugin will start with all devices marked as healthy, but then reacts to the device disappearing by marking it as unhealthy. So since the unhealthy device is one that disappeared, it seems reasonable to use the same distinction when we make a device ready for passthrough, and make it appear for a VM to allocate.

# What isn't included in this PR:
- An implementation of `Allocate` that will allow exact matching between a PCIDevice and a VM. That is coming in a separate PR 

# Related Issues:
- https://github.com/harvester/harvester/issues/2991
- https://github.com/harvester/harvester/issues/3146
- https://github.com/harvester/harvester/issues/3126

### DIFF
--- a/pkg/controller/pcidevice/pcidevice_controller.go
+++ b/pkg/controller/pcidevice/pcidevice_controller.go
@@ -6,6 +6,8 @@ import (
 	"os"
 	"time"
 
+	"github.com/harvester/pcidevices/pkg/iommu"
+
 	v1beta1 "github.com/harvester/pcidevices/pkg/apis/devices.harvesterhci.io/v1beta1"
 	ctl "github.com/harvester/pcidevices/pkg/generated/controllers/devices.harvesterhci.io/v1beta1"
 	"github.com/harvester/pcidevices/pkg/util/nichelper"
@@ -60,7 +62,12 @@ func Register(
 }
 
 func (h Handler) reconcilePCIDevices(nodename string) error {
-	// List all PCI Devices on host
+	// Build up the IOMMU group map
+	iommuGroupPaths, err := iommu.GroupPaths()
+	if err != nil {
+		return err
+	}
+	iommuGroupMap := iommu.GroupMapForPCIDevices(iommuGroupPaths)
 
 	commonLabels := map[string]string{"nodename": nodename} // label
 	var setOfRealPCIAddrs map[string]bool = make(map[string]bool)
@@ -69,9 +76,7 @@ func (h Handler) reconcilePCIDevices(nodename string) error {
 			setOfRealPCIAddrs[dev.Address] = true
 			name := v1beta1.PCIDeviceNameForHostname(dev, nodename)
 			// Check if device is stored
-			var err error
-			var devCR *v1beta1.PCIDevice
-			devCR, err = h.client.Get(name, metav1.GetOptions{})
+			devCR, err := h.client.Get(name, metav1.GetOptions{})
 
 			if err != nil {
 				if apierrors.IsNotFound(err) {
@@ -95,13 +100,14 @@ func (h Handler) reconcilePCIDevices(nodename string) error {
 
 			devCopy := devCR.DeepCopy()
 			// Update only modifies the status, no need to update the main object
-			devCopy.Status.Update(dev, nodename) // update the in-memory CR with the current PCI info
+			devCopy.Status.Update(dev, nodename, iommuGroupMap) // update the in-memory CR with the current PCI info
 			_, err = h.client.UpdateStatus(devCopy)
 			if err != nil {
 				logrus.Errorf("[PCIDeviceController] Failed to update status sub-resource: %v", err)
 				return err
 			}
 		}
+
 	}
 
 	// remove non-existent devices
--- a/pkg/controller/pcideviceclaim/pcideviceclaim_controller.go
+++ b/pkg/controller/pcideviceclaim/pcideviceclaim_controller.go
@@ -9,18 +9,20 @@ import (
 	"time"
 
 	"github.com/harvester/pcidevices/pkg/apis/devices.harvesterhci.io/v1beta1"
+	"github.com/harvester/pcidevices/pkg/deviceplugins"
 	v1beta1gen "github.com/harvester/pcidevices/pkg/generated/controllers/devices.harvesterhci.io/v1beta1"
 	"github.com/sirupsen/logrus"
 	"github.com/spf13/pflag"
 	"github.com/u-root/u-root/pkg/kmodule"
+	kubevirtv1 "kubevirt.io/api/core/v1"
+
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	kubevirtv1 "kubevirt.io/api/core/v1"
 	"kubevirt.io/client-go/kubecli"
 )
 
 const (
-	reconcilePeriod = time.Minute * 1
+	reconcilePeriod = time.Minute * 20
 	vfioPCIDriver   = "vfio-pci"
 	DefaultNS       = "harvester-system"
 	KubevirtCR      = "kubevirt"
@@ -31,10 +33,11 @@ type Controller struct {
 }
 
 type Handler struct {
-	pdcClient  v1beta1gen.PCIDeviceClaimClient
-	pdClient   v1beta1gen.PCIDeviceClient
-	virtClient kubecli.KubevirtClient
-	nodeName   string
+	pdcClient     v1beta1gen.PCIDeviceClaimClient
+	pdClient      v1beta1gen.PCIDeviceClient
+	virtClient    kubecli.KubevirtClient
+	nodeName      string
+	devicePlugins map[string]*deviceplugins.PCIDevicePlugin
 }
 
 func Register(
@@ -43,32 +46,30 @@ func Register(
 	pdClient v1beta1gen.PCIDeviceController,
 ) error {
 	logrus.Info("Registering PCI Device Claims controller")
+	nodename := os.Getenv("NODE_NAME")
 	clientConfig := kubecli.DefaultClientConfig(&pflag.FlagSet{})
 	virtClient, err := kubecli.GetKubevirtClientFromClientConfig(clientConfig)
 	if err != nil {
 		msg := fmt.Sprintf("cannot obtain KubeVirt client: %v", err)
 		return errors.New(msg)
 	}
-	nodename := os.Getenv("NODE_NAME")
+
 	handler := &Handler{
-		pdcClient:  pdcClient,
-		pdClient:   pdClient,
-		virtClient: virtClient,
-		nodeName:   nodename,
+		pdcClient:     pdcClient,
+		pdClient:      pdClient,
+		nodeName:      nodename,
+		virtClient:    virtClient,
+		devicePlugins: make(map[string]*deviceplugins.PCIDevicePlugin),
 	}
+
+	loadVfioDrivers()
 
 	pdcClient.OnRemove(ctx, "PCIDeviceClaimOnRemove", handler.OnRemove)
 	pdcClient.OnChange(ctx, "PCIDeviceClaimReconcile", handler.reconcilePCIDeviceClaims)
-	err = handler.rebindAfterReboot()
-	if err != nil {
-		return err
-	}
 	err = handler.unbindOrphanedPCIDevices()
 	if err != nil {
 		return err
 	}
-	// Load VFIO drivers when controller starts instead of repeatedly in the reconcile loop
-	loadVfioDrivers()
 	return nil
 }
 
@@ -77,10 +78,49 @@ func (h Handler) OnRemove(name string, pdc *v1beta1.PCIDeviceClaim) (*v1beta1.PC
 	if pdc == nil || pdc.DeletionTimestamp == nil || pdc.Spec.NodeName != h.nodeName {
 		return pdc, nil
 	}
-	if pdc == nil {
-		return nil, nil
+
+	// Get PCIDevice for the PCIDeviceClaim
+	pd, err := h.getPCIDeviceForClaim(pdc)
+	if err != nil {
+		logrus.Errorf("Error getting claim's device: %s", err)
+		return pdc, err
 	}
-	return h.attemptToDisablePassthrough(pdc)
+
+	// Get PCIDevice for the PCIDeviceClaim
+	pd, err = h.getPCIDeviceForClaim(pdc)
+	if err != nil {
+		logrus.Errorf("Error getting claim's device: %s", err)
+		return pdc, err
+	}
+
+	// Disable PCI Passthrough by unbinding from the vfio-pci device driver
+	err = h.attemptToDisablePassthrough(pd, pdc)
+	if err != nil {
+		return pdc, err
+	}
+
+	// Find the DevicePlugin
+	resourceName := pd.Status.ResourceName
+	dp := deviceplugins.Find(
+		resourceName,
+		h.devicePlugins,
+	)
+	if dp != nil {
+		err = dp.RemoveDevice(pd, pdc)
+		if err != nil {
+			return pdc, err
+		}
+		// Check if that was the last device, and then shut down the dp
+		time.Sleep(5 * time.Second)
+		if dp.GetCount() == 0 {
+			err := dp.Stop()
+			if err != nil {
+				return pdc, err
+			}
+			delete(h.devicePlugins, resourceName)
+		}
+	}
+	return pdc, nil
 }
 
 func loadVfioDrivers() {
@@ -115,13 +155,9 @@ func bindDeviceToVFIOPCIDriver(pd *v1beta1.PCIDevice) error {
 
 // Enabling passthrough for a PCI Device requires two steps:
 // 1. Bind the device to the vfio-pci driver in the host
-// 2. Permit the device in KubeVirt's Config in the cluster
+// 2. Add device to DevicePlugin so KubeVirt will recognize it
 func (h Handler) enablePassthrough(pd *v1beta1.PCIDevice) error {
 	err := bindDeviceToVFIOPCIDriver(pd)
-	if err != nil {
-		return err
-	}
-	err = h.addHostDeviceToKubeVirt(pd)
 	if err != nil {
 		return err
 	}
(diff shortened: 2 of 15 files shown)


pkg/controller/pcideviceclaim/pcideviceclaim_controller.go
```diff
@@ -343,6 +255,58 @@ func (h Handler) reconcilePCIDeviceClaims(name string, pdc *v1beta1.PCIDeviceCla
 	return pdc, nil
 }
 
+// This function adds the PCIDevice to the device plugin, or creates the device plugin if it doesn't exist
+func (h Handler) addToDevicePlugin(pd *v1beta1.PCIDevice, pdc *v1beta1.PCIDeviceClaim) error {
+	var err error
+	resourceName := pd.Status.ResourceName
+	logrus.Infof("[addToDevicePlugin] finding device plugin %s", resourceName)
+	dp, err := deviceplugins.Find(
```
ibrokethecloud, line 263: nit: the struct already references pointer values, may be should just check if the value is not nil, rather than returning a error?

pkg/deviceplugins/device_manager.go
```diff
@@ -0,0 +1,502 @@
...
+
+	logger := log.DefaultLogger()
+	dev := dpi
+	deviceName := dev.GetDeviceName()
+	logger.Infof("Starting a device plugin for device: %s", deviceName)
+	retries := 0
+
+	backoff := c.backoff
+
+	go func() {
```
ibrokethecloud, line 157: we could use [errgrp](https://pkg.go.dev/golang.org/x/sync/errgroup), and return the error from the go routine. This could be handled by the pcideviceclaim controller, and we could just use the k8s exponential back off rather than writing our own logic

pkg/controller/pcidevice/pcidevice_controller.go
```diff
@@ -61,17 +63,26 @@ func Register(
 
 func (h Handler) reconcilePCIDevices(nodename string) error {
 	// List all PCI Devices on host
+	pci, err := ghw.PCI()
```
ibrokethecloud, line 66: lookup of PCI devices is not needed, as the `Register` method is already handling the same

pkg/controller/pcidevice/pcidevice_controller.go
```diff
@@ -61,17 +63,26 @@ func Register(
...
+	if err != nil {
+		return err
+	}
```
ibrokethecloud, line 69:
```suggestion
``````suggestion
	pci, err := ghw.PCI()
	if err != nil {
		return err
	}
```

pkg/controller/pcidevice/pcidevice_controller.go
```diff
@@ -61,17 +63,26 @@ func Register(
...
+	iommuGroupPaths, err := iommu.GroupPaths()
+	if err != nil {
+		return err
+	}
+	iommuGroupMap := iommu.GroupMapForPCIDevices(iommuGroupPaths)
 
 	commonLabels := map[string]string{"nodename": nodename} // label
 	var setOfRealPCIAddrs map[string]bool = make(map[string]bool)
-	for _, dev := range h.pci.Devices {
+	for _, dev := range pci.Devices {
```
ibrokethecloud, line 79:
```suggestion
	for _, dev := range h.pci.Devices {
``````suggestion
	for _, dev := range pci.Devices {
```

pkg/controller/pcidevice/pcidevice_controller.go
```diff
@@ -61,17 +63,26 @@ func Register(
...
 		if !containsString(h.skipAddresses, dev.Address) {
 			setOfRealPCIAddrs[dev.Address] = true
 			name := v1beta1.PCIDeviceNameForHostname(dev, nodename)
 			// Check if device is stored
 			var err error
```
ibrokethecloud, line 84:
```suggestion
``````suggestion
			var err error
```
not needed as we are already initialising err in the next line

pkg/controller/pcidevice/pcidevice_controller.go
tlehman: change it to the `h.pci` right?

pkg/controller/pcidevice/pcidevice_controller.go
tlehman: Shouldn't this be `pci := h.pci`?

pkg/controller/pcidevice/pcidevice_controller.go
ibrokethecloud: 👍

ibrokethecloud: actually this is not needed at all. since we already are reconcilling and storing state of pci devices in the handler. So we need to range over h.pci.

pkg/controller/pcideviceclaim/pcideviceclaim_controller.go
```diff
@@ -335,45 +290,139 @@ func (h Handler) reconcilePCIDeviceClaims(name string, pdc *v1beta1.PCIDeviceCla
...
+			break
+		}
+	}
+	if !devPermitted {
+		vendorId := pd.Status.VendorId
+		deviceId := pd.Status.DeviceId
+		devToPermit := kubevirtv1.PciHostDevice{
+			PCIVendorSelector:        fmt.Sprintf("%s:%s", vendorId, deviceId),
+			ResourceName:             resourceName,
+			ExternalResourceProvider: false,
```
ibrokethecloud, line 361: for the plugin to be leverage the ExternalResourceProvider should be true

```suggestion
			ExternalResourceProvider: true,
```

pkg/controller/pcideviceclaim/pcideviceclaim_controller.go
```diff
@@ -335,45 +290,139 @@ func (h Handler) reconcilePCIDeviceClaims(name string, pdc *v1beta1.PCIDeviceCla
 		return pdc, nil
 	}
 
-	if pdc.Spec.NodeName == h.nodeName && !pdc.Status.PassthroughEnabled {
-		newPdc, err := h.attemptToEnablePassthrough(pdc)
-		return newPdc, err
+	if pdc.Spec.NodeName != h.nodeName || pdc.Status.PassthroughEnabled {
```
ibrokethecloud, line 293: currently if a pcideviceclaim is already enabled, then we dont start the plugin due to this check.

tlehman: Fixed in a new commit. I moved the node check to the beginning of the reconcile function, then only returned if `PDC.passthroughEnabled && dp != nil`

pkg/deviceplugins/device_manager.go
```diff
@@ -0,0 +1,490 @@
...
+		}
+		containerResponse.Devices = deviceSpecs
+		envVar := make(map[string]string)
+		envVar[resourceNameEnvVar] = strings.Join(allocatedDevices, ",")
+
+		containerResponse.Envs = envVar
+		resp.ContainerResponses = append(resp.ContainerResponses, containerResponse)
+	}
+	return resp, nil
+}
```
ibrokethecloud, line 277:
```suggestion
func (dpi *PCIDevicePlugin) Allocate(_ context.Context, r *pluginapi.AllocateRequest) (*pluginapi.AllocateResponse, error) {
	logrus.Debugf("Allocate request %s", r.String())
	resourceNameEnvVar := util.ResourceNameToEnvVar(PCIResourcePrefix, dpi.resourceName)
	allocatedDevices := []string{}
	resp := new(pluginapi.AllocateResponse)
	containerResponse := new(pluginapi.ContainerAllocateResponse)

	for _, request := range r.ContainerRequests {
		deviceSpecs := make([]*pluginapi.DeviceSpec, 0)
		for _, devID := range request.DevicesIDs {
			// translate device's iommu group to its pci address
			logrus.Debugf("looking up deviceID %s in map %v", devID, dpi.iommuToPCIMap)
			iommuGroup, exist := dpi.iommuToPCIMap[devID] // not finding device ids
			if exist {
				// if device exists, check if there other devices
				// in the same iommuGroup, and append these too
				allocatedDevices = append(allocatedDevices, devID)
				for devPCIAddress, ig := range dpi.iommuToPCIMap {
					if ig == iommuGroup {
						allocatedDevices = append(allocatedDevices, devPCIAddress)
					}
				}
				deviceSpecs = append(deviceSpecs, formatVFIODeviceSpecs(iommuGroup)...)
			} else {
				continue // break execution of loop as we are not handling this device
			}
		}
		containerResponse.Devices = deviceSpecs
		envVar := make(map[string]string)
		envVar[resourceNameEnvVar] = strings.Join(allocatedDevices, ",")

		containerResponse.Envs = envVar
		resp.ContainerResponses = append(resp.ContainerResponses, containerResponse)
	}
	logrus.Debugf("Allocate response %v", resp)
	return resp, nil
}
```
Made some changes to Allocate to return correct iommuGroup, and also added additional devices part of the same iommuGroup.
This addresses the error where kubevirt cant create host-devices

pkg/deviceplugins/device_manager.go
```diff
@@ -0,0 +1,490 @@
...
+			stopChan: make(chan struct{}),
+			backoff:  defaultBackoffTime,
+		},
+	}
+	return dpi
+}
+
+func constructDPIdevices(pciDevices []*PCIDevice, iommuToPCIMap map[string]string) (devs []*pluginapi.Device) {
+	for _, pciDevice := range pciDevices {
+		iommuToPCIMap[pciDevice.iommuGroup] = pciDevice.pciAddress
```
ibrokethecloud, line 132:
```suggestion
		iommuToPCIMap[pciDevice.pciAddress] = pciDevice.iommuGroup
```
swapped struct of map to make it easy to lookup iommuGroup from the device address

pkg/deviceplugins/deviceplugin.go
```diff
@@ -0,0 +1,80 @@
...
+	claim *v1beta1.PCIDeviceClaim,
+	pdsWithSameResourceName []*v1beta1.PCIDevice,
+) *PCIDevicePlugin {
+	// Check if there are any PCIDevicePlugins with that resourceName
+	pcidevs := []*PCIDevice{}
+	for _, pd := range pdsWithSameResourceName {
+		pcidevs = append(pcidevs, &PCIDevice{
+			pciID:      pd.Status.Address,
+			driver:     pd.Status.KernelDriverInUse,
+			pciAddress: pd.Status.Address, // this redundancy is here to distinguish between the ID and the PCI Address. They have the same value but mean different things
```
ibrokethecloud, line 51:
```suggestion
			iommuGroup: pd.Status.IOMMUGroup,
```
Added initialisation for iommoGroup in the pcidevices. This was causing issues with lookup of iommoGroup during allocate calls.

pkg/controller/pcideviceclaim/pcideviceclaim_controller.go
```diff
@@ -331,49 +286,144 @@ func (h Handler) rebindAfterReboot() error {
 
 func (h Handler) reconcilePCIDeviceClaims(name string, pdc *v1beta1.PCIDeviceClaim) (*v1beta1.PCIDeviceClaim, error) {
 
-	if pdc == nil || pdc.DeletionTimestamp != nil {
+	if pdc == nil || pdc.DeletionTimestamp != nil || (pdc.Spec.NodeName != h.nodeName) {
```
ibrokethecloud, line 289:
```suggestion
	if pdc.Spec.NodeName != h.nodeName {
```
dropped check to see if device passthrough is already enabled, as we still need to start the plugin if needed.

pkg/controller/pcideviceclaim/pcideviceclaim_controller.go
```diff
@@ -331,49 +286,144 @@ func (h Handler) rebindAfterReboot() error {
...
 	}
 
-	if pdc.Spec.NodeName == h.nodeName && !pdc.Status.PassthroughEnabled {
-		newPdc, err := h.attemptToEnablePassthrough(pdc)
-		return newPdc, err
+	// Get the PCIDevice object for the PCIDeviceClaim
+	pd, err := h.getPCIDeviceForClaim(pdc)
+	if pd == nil {
+		return pdc, err
+	}
```
ibrokethecloud, line 297:
```suggestion
	if err != nil {
		return pdc, fmt.Errorf("error enabling device passthrough: %v", err)
	}
```
check and handle error while attempting passthrough

pkg/controller/pcideviceclaim/pcideviceclaim_controller.go
```diff
@@ -331,49 +286,144 @@ func (h Handler) rebindAfterReboot() error {
...
+	h.permitHostDeviceInKubeVirt(pd)
+
+	// If passthrough is already enabled and there's a deviceplugin, then we can return
+	if pdc.Status.PassthroughEnabled && (dp != nil){
+		return pdc, nil
 	}
 
-	return pdc, nil
+	// Enable PCI Passthrough on the device by binding it to vfio-pci driver
+	newPdc, err := h.attemptToEnablePassthrough(pd, pdc)
```
ibrokethecloud, line 314:
```suggestion
	if err := h.permitHostDeviceInKubeVirt(pd); err != nil {
		return pdc, fmt.Errorf("error updating kubevirt CR: %v", err)
	}
```
added error handling to ensure this is retried in case of issues.

pkg/controller/pcideviceclaim/pcideviceclaim_controller.go
```diff
@@ -331,49 +286,144 @@ func (h Handler) rebindAfterReboot() error {
...
+			break
+		}
+	}
+	if !devPermitted {
+		vendorId := pd.Status.VendorId
+		deviceId := pd.Status.DeviceId
+		devToPermit := kubevirtv1.PciHostDevice{
+			PCIVendorSelector:        fmt.Sprintf("%s:%s", vendorId, deviceId),
+			ResourceName:             resourceName,
+			ExternalResourceProvider: true,
```
ibrokethecloud, line 362:
```suggestion
			ExternalResourceProvider: true,
```
needs to be true, to force kubevirt to use pcidevice plugin

pkg/controller/pcideviceclaim/pcideviceclaim_controller.go
```diff
@@ -115,30 +153,22 @@ func bindDeviceToVFIOPCIDriver(pd *v1beta1.PCIDevice) error {
 
 // Enabling passthrough for a PCI Device requires two steps:
 // 1. Bind the device to the vfio-pci driver in the host
-// 2. Permit the device in KubeVirt's Config in the cluster
-func (h Handler) enablePassthrough(pd *v1beta1.PCIDevice) error {
+// 2. Add device to DevicePlugin so KubeVirt will recognize it
+func (h Handler) enablePassthrough(pd *v1beta1.PCIDevice, pdc *v1beta1.PCIDeviceClaim) error {
```
ibrokethecloud, line 157:
```suggestion
func (h Handler) enablePassthrough(pd *v1beta1.PCIDevice) error {
```
PCIDeviceClaim object is not needed.

pkg/controller/pcideviceclaim/pcideviceclaim_controller.go
```diff
@@ -312,7 +267,7 @@ func (h Handler) rebindAfterReboot() error {
 		}
 
 		// Enable Passthrough on the device
-		err = h.enablePassthrough(pd)
+		err = h.enablePassthrough(pd, pdcCopy)
```
ibrokethecloud, line 270:
```suggestion
		err = h.enablePassthrough(pd)
```

updated to reflect the new method signature

pkg/controller/pcideviceclaim/pcideviceclaim_controller.go
ibrokethecloud:
```suggestion
	pdcCopy := pdc.DeepCopy()
	pdcCopy.Status.KernelDriverToUnbind = pd.Status.KernelDriverInUse
	vfioDriverEnabled := pd.Status.KernelDriverInUse == vfioPCIDriver // it is possible that device was enabled however pcideviceclaim status updated failed

	if !vfioDriverEnabled {
		logrus.Infof("Enabling passthrough for PDC: %s", pdc.Name)
		// Only unbind from driver is a driver is currently in use
		if strings.TrimSpace(pd.Status.KernelDriverInUse) != "" {
			err := unbindDeviceFromDriver(pd.Status.Address, pd.Status.KernelDriverInUse)
			if err != nil {
				pdcCopy.Status.PassthroughEnabled = false
			}
		}
		// Enable PCI Passthrough by binding the device to the vfio-pci driver
		err := h.enablePassthrough(pd)
		if err != nil {
			return pdc, err
		}
	}

	if !pdcCopy.Status.PassthroughEnabled {
		pdcCopy.Status.PassthroughEnabled = true
		return h.pdcClient.UpdateStatus(pdcCopy)
	}

	return pdc, nil

```

minor restructure of reconcile logic. this handles issues where the vfio driver was enable, however due to conflicts, the update to pcideviceObj.Status.Passthrough enabled failed.

czadikem:
How can I try this on harvester?

tlehman:
@czadikem It will be included in Harvester 1.1.2 (not released yet), but if you want to try it sooner, install Harvester 1.1.1, enable the pcidevices addon, and then edit the daemonset to use image `tlehman/pcidevices:dev`

How to edit the daemonset:
```
kubectl -n harvester-system edit ds harvester-pcidevices-controller
```

czadikem:
Thank you I will be sure to try this.

GopeshKh:
@tlehman : Thank you for making this update. However, I noticed that this change does not appear in the 9c9e0960 Release v1.2.1. Could you confirm if these changes were included in the Release v1.2.1?

connorkuehl:
@GopeshKh The PCI Passthrough functionality is packaged in Harvester as an addon[1] which is not enabled by default.

On your Harvester dashboard, you can go to **Advanced** > **Addons**. Then click on the 3 vertical dots for the pcidevices-controller addon > **Enable**.

It'll take a few minutes for the addon to deploy and enumerate the available PCI devices.

If you run into any issues please feel free to file a new GitHub issue or find us in the harvester room on the Rancher User's Slack 🙂 

[1] https://docs.harvesterhci.io/v1.2/advanced/addons/pcidevices

ibrokethecloud:
the latest plugins changes are available in 1.2.1, where we fixed a few more bugs. If you still run into issues post upgrade, please provide us a support bundle.

PR #41, opened 2023-02-20
Skip cluster networks

ibrokethecloud:
Related to: https://github.com/harvester/harvester/issues/3199

Based on feedback from QA, additional logic has been added to skip interfaces from the node, which may be involved in cluster networks.

In addition, a new `tidy` target has been added for dapper to run `goimports-reviser`

This should make it easy to format code across multiple development platforms.


main.go
```diff
@@ -6,6 +6,7 @@ import (
 	"os"
 
 	"github.com/harvester/pcidevices/pkg/controller/pcidevice"
+	ctlcore "github.com/rancher/wrangler/pkg/generated/controllers/core"
```
futuretea, line 9: need to be grouped

scripts/tidy
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
+set -e
+
+export PATH=$PATH:/root/go/bin
+cd $(dirname $0)/..
+for i in "./pkg/codegen" "./pkg/controller" "./pkg/crd" "./pkg/deviceplugins" "./pkg/util" "./pkg/webhook"
```
futuretea, line 6: main.go is also needed

pkg/controller/pcidevice/pcidevice_controller.go
```diff
@@ -25,20 +30,30 @@ const (
 type Handler struct {
 	client        ctl.PCIDeviceClient
 	pci           *ghw.PCIInfo
+	nodeCache     ctlcorev1.NodeCache
+	networkCache  ctlnetworkv1beta1.VlanConfigCache
```
futuretea, line 34: s/networkCache/vlanConfigCache/

pkg/controller/pcidevice/pcidevice_controller.go
```diff
@@ -25,20 +30,30 @@ const (
...
 	handler := &Handler{
-		client: pd,
+		client:       pd,
+		nodeCache:    coreFactory.Core().V1().Node().Cache(),
+		networkCache: networkFactory.Network().V1beta1().VlanConfig().Cache(),
 	}
 
 	nodename := os.Getenv("NODE_NAME")
+
+	if err := start.All(ctx, 1, coreFactory, networkFactory); err != nil {
```
futuretea, line 53: can we do this in https://github.com/harvester/pcidevices/blob/0a1f7c2b535e5729aa05dcdb47a28b5a0cafb83b/main.go#L123?

ibrokethecloud: the pcidevice register is a blocking call, hence the need to run it within  the register method.

ibrokethecloud:
Trivy scanning is currently failing due to inabiity to detect the sem version for rancher/rancher dependencies.

I have already updated go.mod to use `rancher/rancher` commit matching tag v2.6.10

futuretea (approved):
LGTM, thanks

pkg/util/nichelper/helper.go
```diff
@@ -67,3 +80,45 @@ func IdentifyManagementNIC() ([]string, error) {
...
+		if ok {
+			nicsList = append(nicsList, v.Spec.Uplink.NICs...)
+		}
+	}
+	return nicsList, nil
+}
+
+// currentNodeMatchesSelector will use the label selectors from VlanConfig to identify if node is
+// in the matching the VlanConfig
+func currentNodeMatchesSelector(nodeName string, nodeCache ctlcorev1.NodeCache, vlanConfigLabels map[string]string) (bool, error) {
```
yaocw2020, line 106: The vlanconfig has an annotation network.harvesterhci.io/matched-nodes which records the matching nodes. The harvester network webhook adds the annotation automatically. It would be more convenient to use the annotation.

ibrokethecloud: does the webhook update the annotation when nodes matching label selector are added to the cluster?

yaocw2020: Yes.

yaocw2020 (approved):
LGTM, thanks.

PR #46, opened 2023-03-01
clean up startup issues

ibrokethecloud:
fixed up registration of controllers

**Problem:**

node controller registration was incorrect which caused node cleanup controller to not be triggered during removal of node objects.

**Solution:**

**Related Issue:**

**Test plan:**

### DIFF
--- a/main.go
+++ b/main.go
@@ -1,28 +1,23 @@
 package main
 
 import (
-	"context"
 	"fmt"
 	"os"
 
 	harvesternetworkv1beta1 "github.com/harvester/harvester-network-controller/pkg/apis/network.harvesterhci.io/v1beta1"
 	ctlnetwork "github.com/harvester/harvester-network-controller/pkg/generated/controllers/network.harvesterhci.io"
-	"github.com/rancher/lasso/pkg/cache"
-	"github.com/rancher/lasso/pkg/client"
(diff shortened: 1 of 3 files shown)


main.go
```diff
@@ -137,27 +120,18 @@ func run(kubeConfig string) error {
 	if err != nil {
 		return fmt.Errorf("error building network controllers: %v", err)
 	}
-	pdCtl := factory.Devices().V1beta1().PCIDevice()
-	pdcCtl := factory.Devices().V1beta1().PCIDeviceClaim()
-	nodeCtl := corecontrollers.Core().V1().Node()
+	pdCtl := pciFactory.Devices().V1beta1().PCIDevice()
+	pdcCtl := pciFactory.Devices().V1beta1().PCIDeviceClaim()
+	nodeCtl := coreFactory.Core().V1().Node()
```
tlehman, line 125: This is better naming ✅

tlehman (approved):
LGTM

guangbochen (approved):
LGTM, thanks.

main.go
ibrokethecloud: yeah that is what was causing the confusion 😆

bk201 (approved):

PR #50, opened 2023-03-08
Security upgrades

tlehman:
**Problem:**
Upgrading vulnerable packages based on the security scan done in https://github.com/harvester/security/issues/19

**Solution:**
- **github.com/rancher/wrangler** 1.0.1 -> 1.1.0
- **kubevirt.io/kubevirt** -> 55.1
- **golang.org/x/net** -> 0.7.0 (highest version covered all three CVEs)
- For **github.com/u-root/u-root**, I verified that the fix was is 0.11.0, but when I attempt to upgrade to 0.11.0, the `go mod tidy` step downgrades kubevirt to 0.17.4. Since we are only using u-root/pkg/kmodule, in [one place](https://github.com/tlehman/pcidevices/blob/master/pkg/controller/pcideviceclaim/pcideviceclaim_controller.go#L130), none of the three CVEs flagged for this version apply to our usage of the library, so it is safe for us to leave it alone for now.
  - [[CVE-2020-7665](https://nvd.nist.gov/vuln/detail/CVE-2020-7665)](https://nvd.nist.gov/vuln/detail/CVE-2020-7665) pkg/uzip **path traversal attack**                                                                                             
  - [[CVE-2020-7666](https://nvd.nist.gov/vuln/detail/CVE-2020-7666)](https://nvd.nist.gov/vuln/detail/CVE-2020-7666) pkg/cpio **path traversal and symlink attacks**
  - [[CVE-2020-7669](https://nvd.nist.gov/vuln/detail/CVE-2020-7669)](https://nvd.nist.gov/vuln/detail/CVE-2020-7669) pkg/tarutil **path traversal and symlink attacks**
    - https://github.com/u-root/u-root/pull/1817 "zipslip"
(Notice how pkg/kmodule is not implicated, also these vulnerabilities are only exploitable with a malicious, user-crafted path input. We use a hard-coded path in a non-affected package, so these vulnerabilities can be safely dismissed in our case).

**Related Issue:**
https://github.com/harvester/security/issues/19

**Test plan:**
Make sure none of the 1.1.2 functionality related to PCI devices is broken by this

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -13,18 +13,18 @@ require (
 	github.com/onsi/ginkgo/v2 v2.1.6
 	github.com/onsi/gomega v1.20.1
 	github.com/rancher/dynamiclistener v0.3.5
-	github.com/rancher/lasso v0.0.0-20220628160937-749b3397db38
-	github.com/rancher/wrangler v1.0.2
+	github.com/rancher/lasso v0.0.0-20221227210133-6ea88ca2fbcc
+	github.com/rancher/wrangler v1.1.0
 	github.com/sirupsen/logrus v1.9.0
-	github.com/stretchr/testify v1.8.0
+	github.com/stretchr/testify v1.8.1
 	github.com/u-root/u-root v7.0.0+incompatible
 	github.com/urfave/cli/v2 v2.11.1
 	github.com/vishvananda/netlink v1.2.1-beta.2
 	google.golang.org/grpc v1.48.0
-	k8s.io/api v0.24.7
-	k8s.io/apimachinery v0.24.7
+	k8s.io/api v0.25.4
+	k8s.io/apimachinery v0.25.4
 	k8s.io/client-go v12.0.0+incompatible
-	k8s.io/kube-aggregator v0.24.0
+	k8s.io/kube-aggregator v0.25.4
 	kubevirt.io/client-go v0.54.0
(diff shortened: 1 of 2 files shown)


go.mod
```diff
@@ -109,8 +109,8 @@ require (
 	gopkg.in/inf.v0 v0.9.1 // indirect
 	gopkg.in/yaml.v2 v2.4.0 // indirect
 	gopkg.in/yaml.v3 v3.0.1 // indirect
-	k8s.io/apiextensions-apiserver v0.24.3 // indirect
-	k8s.io/code-generator v0.24.3 // indirect
+	k8s.io/apiextensions-apiserver v0.25.4 // indirect
+	k8s.io/code-generator v0.25.4 // indirect
```
guangbochen, line 113: prefer to patch the indirect import via the `repalce` section.

tlehman: In the [vm-import-controller](https://github.com/harvester/vm-import-controller/pull/13/files#:~:text=k8s.io/code%2Dgenerator%20v0.25.4%20//%20indirect) this was the same approach used. Unless we have a version conflict or something, I don't think it's necessary to do a manual replace

guangbochen (approved):
LGTM, thanks.

bk201 (approved):
Thanks!

PR #51, opened 2023-03-10
Fixes for iommu grouping and pcidevice claim bind/unbind

ibrokethecloud:

**Problem:**

PR addresses two issues:
* bind/unbind reports device busy. This is likely due to lack of leader election logic, where node X may process a PCIDeviceClaim for node Y. In this scenario it will just ignore object, and wrangler ends up removing the finalizer. however the actual unbinding/binding may not have been performed. requeuing such objects allows us to ensure device is rebound to correct driver. additional checks have been added to ensure that device is correct bound and available in driver tree before operation is reconcilled successfully.
* in cases where all pcidevices of an iommu group are already present in a VM, no additional devices should be added. there was a bug where we were not checking if other devices from iommu group are already in the VM definition, resulting in devices being duplicated in the patch generated by mutating webhook. this has been correct as well. related issue: https://github.com/harvester/harvester/issues/3631
* 
**Solution:**

**Related Issue:**

**Test plan:**

### DIFF
--- a/pkg/controller/pcideviceclaim/pcideviceclaim_controller.go
+++ b/pkg/controller/pcideviceclaim/pcideviceclaim_controller.go
@@ -82,8 +82,13 @@ func Register(
 
 // When a PCIDeviceClaim is removed, we need to unbind the device from the vfio-pci driver
 func (h *Handler) OnRemove(name string, pdc *v1beta1.PCIDeviceClaim) (*v1beta1.PCIDeviceClaim, error) {
-	if pdc == nil || pdc.DeletionTimestamp == nil || pdc.Spec.NodeName != h.nodeName {
+	if pdc == nil || pdc.DeletionTimestamp == nil {
 		return pdc, nil
+	}
+
+	// need to requeue object to ensure correct node picks up and cleanup/rebind is executed
+	if pdc.Spec.NodeName != h.nodeName {
+		return pdc, fmt.Errorf("requeue object to ensure its picked up by correct node")
 	}
 
 	// Get PCIDevice for the PCIDeviceClaim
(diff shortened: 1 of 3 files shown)


tlehman (approved):
LGTM, also I just tested this and it resolves the reboot issue

pkg/controller/pcideviceclaim/pcideviceclaim_controller.go
```diff
@@ -82,10 +83,15 @@ func Register(
...
 // When a PCIDeviceClaim is removed, we need to unbind the device from the vfio-pci driver
 func (h *Handler) OnRemove(name string, pdc *v1beta1.PCIDeviceClaim) (*v1beta1.PCIDeviceClaim, error) {
-	if pdc == nil || pdc.DeletionTimestamp == nil || pdc.Spec.NodeName != h.nodeName {
+	if pdc == nil || pdc.DeletionTimestamp == nil {
 		return pdc, nil
 	}
 
+	// need to requeue object to ensure correct node picks up and cleanup/rebind is executed
+	if pdc.Spec.NodeName != h.nodeName {
+		return pdc, fmt.Errorf("requeue object to ensure its picked up by correct node")
```
guangbochen, line 92: prefer to call enqueue method instead.

pkg/controller/pcideviceclaim/pcideviceclaim_controller.go
```diff
@@ -134,20 +140,44 @@ func loadVfioDrivers() {
...
+
+	vendorId := pd.Status.VendorId
+	deviceId := pd.Status.DeviceId
+	var id string = fmt.Sprintf("%s %s", vendorId, deviceId)
+	logrus.Infof("Binding device %s [%s] to vfio-pci", pd.Name, id)
+
+	newIdFile, err := os.OpenFile("/sys/bus/pci/drivers/vfio-pci/new_id", os.O_WRONLY, 0200)
+	if err != nil {
+		logrus.Errorf("Error opening new_id file")
+		return err
```
guangbochen, line 155: can we just return the error with the above message since the controller should print the message directly?

pkg/controller/pcideviceclaim/pcideviceclaim_controller.go
```diff
@@ -134,20 +140,44 @@ func loadVfioDrivers() {
...
+	if err != nil && !os.IsExist(err) {
+		logrus.Errorf("Error writing to new_id file: %s", err)
+		return err
+	}
 
+	logrus.Infof("Binding device %s vfio-pci", pd.Status.Address)
 	file, err := os.OpenFile("/sys/bus/pci/drivers/vfio-pci/bind", os.O_WRONLY, 0200)
 	if err != nil {
-		logrus.Errorf("Error opening new_id file: %s", err)
+		logrus.Errorf("Error opening bind file: %s", err)
```
guangbochen, line 167: ditto

pkg/controller/pcideviceclaim/pcideviceclaim_controller.go
```diff
@@ -134,20 +140,44 @@ func loadVfioDrivers() {
...
 	if err != nil {
-		logrus.Errorf("Error writing to new_id file: %s", err)
+		logrus.Errorf("Error writing to bind file: %s", err)
 		file.Close()
 		return err
 	}
 	file.Close()
+
+	if !deviceBoundToDriver(vfioPCIDriverPath, pd.Status.Address) {
+		return fmt.Errorf("no device %s found at /sys/bus/pci/drivers/vfio-pci", pd.Status.Address)
```
guangbochen, line 179: use `bindDeviceToVFIOPCIDriver` instead.

ibrokethecloud: the error will result in object being requeued and falling through the reconcile loop

bk201 (approved):
LGTM, thanks!

guangbochen (approved):
LGTM, thanks.

PR #52, opened 2023-04-11
change how network links are looked up without hostNetwork

ibrokethecloud:

**Problem:**

PR changes how netlink library is used to access the host network namespace, without use of hostNetwork.

As a result of this change, the mutating webhook pod is no longer exposed on the host network.

**Solution:**

**Related Issue:**

**Test plan:**

### DIFF
--- a/Dockerfile.dapper
+++ b/Dockerfile.dapper
@@ -22,6 +22,7 @@ ENV DAPPER_ENV REPO TAG DRONE_TAG
 ENV DAPPER_SOURCE /go/src/github.com/harvester/pcidevices/
 ENV DAPPER_OUTPUT ./bin ./dist ./pkg
 ENV DAPPER_DOCKER_SOCKET true
+ENV DAPPER_RUN_ARGS "-v /proc:/host/proc --privileged"
 ENV HOME ${DAPPER_SOURCE}
 WORKDIR ${DAPPER_SOURCE}
 
--- a/pkg/util/nichelper/helper.go
+++ b/pkg/util/nichelper/helper.go
@@ -20,6 +20,46 @@ const (
 )
 
 func IdentifyHarvesterManagedNIC(nodeName string, nodeCache ctlcorev1.NodeCache, vlanConfigCache ctlnetworkv1beta1.VlanConfigCache) ([]string, error) {
+	var skipInterfaces []string
+	managementInterfaces, err := IdentifyManagementNics()
+	if err != nil {
+		return nil, err
+	}
+
+	skipInterfaces = append(skipInterfaces, managementInterfaces...)
+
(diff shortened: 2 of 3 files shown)


pkg/util/nichelper/helper.go
```diff
@@ -19,7 +20,17 @@ const (
...
+	hostProcessNS, err := netns.GetFromPath("/host/proc/1/ns/net")
+	if err != nil {
+		return nil, fmt.Errorf("error fetching host network namespace: %v", err)
+	}
+
+	handler, err := netlink.NewHandleAt(hostProcessNS)
+	if err != nil {
+		return nil, fmt.Errorf("error generating handler for host network namespace: %v", err)
+	}
+
```
guangbochen, line 72: ditto, the handler should be closed on `defer`

pkg/util/nichelper/helper.go
```diff
@@ -78,6 +57,52 @@ func IdentifyHarvesterManagedNIC(nodeName string, nodeCache ctlcorev1.NodeCache,
...
+	// skipInterfaces contains names of interfaces using in the default harvester bonding interfaces
+	// these are to be used by the PCIDevices controller to skip said devices
+	var skipInterfaces []string
+	for _, i := range masterBondedIndexes {
+		for _, l := range link {
+			// check helps skip over cases when mgmt-br is also being used for vm networks
+			// in which case mgmt-bo is also pointing to mgmt-br
+			if l.Attrs().Slave != nil {
+				if l.Attrs().MasterIndex == i {
```
guangbochen, line 96: may use `if link.Attrs().Slave != nil && link.Attrs().MasterIndex == i {` instead.

Dockerfile.dapper
```diff
@@ -22,6 +22,7 @@ ENV DAPPER_ENV REPO TAG DRONE_TAG
 ENV DAPPER_SOURCE /go/src/github.com/harvester/pcidevices/
 ENV DAPPER_OUTPUT ./bin ./dist ./pkg
 ENV DAPPER_DOCKER_SOCKET true
+ENV DAPPER_RUN_ARGS "-v /proc:/host/proc --privileged"
```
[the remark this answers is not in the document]
ibrokethecloud: This is only needed for retrofitting changes into SRIOV PR, which has additional unit / integration tests which attempt to query the host nics.

guangbochen (approved):
LGTM, and can we add related GH issue for it, thanks.

starbops (approved):
LGTM, thanks for the fix!

futuretea (approved):
LGTM, thanks

tlehman (approved):
Big improvement. LGTM

PR #55, opened 2023-05-23
fix validation errors and enable additional checks in CI

ibrokethecloud:

**Problem:**
Include additional ci checks and fix reconcile issues with device plugins.

Related to: https://github.com/harvester/harvester/issues/3524

**Solution:**

**Related Issue:**

**Test plan:**


pkg/controller/pcideviceclaim/pcideviceclaim_controller.go
```diff
@@ -107,6 +108,8 @@ func (h *Handler) OnRemove(name string, pdc *v1beta1.PCIDeviceClaim) (*v1beta1.P
 		return pdc, err
 	}
 
+	lock.Lock()
```
guangbochen, line 111: any reasons why the lock is requried on the remove method?

pkg/controller/pcideviceclaim/pcideviceclaim_controller.go
```diff
@@ -290,15 +287,27 @@ func (h *Handler) reconcilePCIDeviceClaims(name string, pdc *v1beta1.PCIDeviceCl
...
```
guangbochen, line 290: ditto

pkg/deviceplugins/device_manager.go
```diff
@@ -148,91 +148,95 @@ func constructDPIdevices(pciDevices []*PCIDevice, iommuToPCIMap map[string]strin
...
-	sock, err := net.Listen("unix", dpi.socketPath)
+	sock, err := net.Listen("unix", dp.socketPath)
 	if err != nil {
 		return fmt.Errorf("error creating GRPC server socket: %v", err)
 	}
 
-	dpi.server = grpc.NewServer([]grpc.ServerOption{}...)
-	defer dpi.stopDevicePlugin()
+	dp.server = grpc.NewServer([]grpc.ServerOption{}...)
+	//defer dpi.stopDevicePlugin()
```
guangbochen, line 184: remove this line if not required.

pkg/deviceplugins/common.go
```diff
@@ -127,20 +129,17 @@ func waitForGRPCServer(socketPath string, timeout time.Duration) error {
 
 // dial establishes the gRPC communication with the registered device plugin.
 func gRPCConnect(socketPath string, timeout time.Duration) (*grpc.ClientConn, error) {
-	c, err := grpc.Dial(socketPath,
-		grpc.WithInsecure(),
+	ctx, cancel := context.WithTimeout(context.Background(), timeout)
```
guangbochen, line 132: any preference to initialize a new context rather than pass through the main ctx via the handler?

guangbochen (commented):
btw, prefer to bump to golang 1.19 if it is a manageable change, thanks.

pkg/controller/pcideviceclaim/pcideviceclaim_controller.go
ibrokethecloud: The controller is running two threads by default.

The locking was introduced to ensure that device plugin updates are serialized.

The remove is also managed by a lock, as the plugin is deleted if there are no more pcideviceclaims for the device type and we need to ensure that plugin is working until the last deviceclaim is deleted.

pkg/controller/pcideviceclaim/pcideviceclaim_controller.go
ibrokethecloud: same as above..

ibrokethecloud:
To test device plugin fix can only be tested with multiple devices of the same type.

Easiest option would be to create VF's on a nic or use the SriovNetworkDevices crd.

* create 4 VF's
* enable pcidevice passthrough for all 4 VF's.
* check node status `kubectl get nodes -o yaml | yq .status.allocatable`, user should see 4 devices available for the VF's just created.
* delete pcideviceclaim pod or reboot node.
* check node status `kubectl get nodes -o yaml | yq .status.allocatable`, without this fix there is a high chance of only seeing 3 devices available for the VF's created from step above.
* build a new image with fixes, and update pcidevices-controller
* once controller initialises `kubectl get nodes -o yaml | yq .status.allocatable` should reflect the right number of VF's.

pkg/controller/pcideviceclaim/pcideviceclaim_controller.go
guangbochen: Adding this as a code comment will be helpful, thanks.

pkg/deviceplugins/common.go
ibrokethecloud: logic has been updated to store the main context into pcideviceclaim handler, and this is in turn stored in pcideviceplugin, for use when the grpc connection is initialized.

pkg/controller/pcideviceclaim/pcideviceclaim_controller.go
ibrokethecloud: done.

guangbochen (approved):
LGTM, thanks.

pkg/webhook/vm.go
```diff
@@ -88,7 +88,7 @@ func (vm *vmPCIMutator) Update(request *types.Request, oldObj runtime.Object, ne
 
 // generatePatch is a common method used by create and update calls to generate a patch operation for VM
 func (vm *vmPCIMutator) generatePatch(vmObj *kubevirtv1.VirtualMachine) (types.PatchOps, error) {
-	var pciDevicesInVM []string
+	pciDevicesInVM := make([]string, len(vmObj.Spec.Template.Spec.Domain.Devices.HostDevices))
```
futuretea, line 91: pciDevicesInVM := make([]string, 0, len(vmObj.Spec.Template.Spec.Domain.Devices.HostDevices))

ibrokethecloud: fixed

futuretea (approved):
LGTM, thanks.

bk201 (approved):

WebberHuang1118:
https://github.com/ibrokethecloud/pcidevices/blob/b6653138864acb5fe4842f2d0864eb71d6c0da2e/pkg/controller/pcidevice/pcidevice_controller.go#L69
maybe not directly related to this PR, but it logs with wrong information?

ibrokethecloud:
good spot @WebberHuang1118. fixed it.

WebberHuang1118 (approved):
LGTM, thanks

PR #60, opened 2023-11-23
support for sriovgpu and vgpu management

ibrokethecloud:

**Problem:**

PR introduces the initial work to allow management of NVIDIA GPU's capable of sriov vgpus.

**Solution:**

**Related Issue:**
https://github.com/harvester/harvester/issues/2764
**Test plan:**


scydas:
Hi，will there be support for sriov vgpu?

pkg/util/executor/interface.go
```diff
@@ -16,8 +20,20 @@ func NewLocalExecutor(envVars []string) *LocalExecutor {
 	}
 }
 
+const (
+	localExecutorAppend = "/host"
```
tserong, line 24: This should probably be called something like `localExecutorPrefix` (vs. "append", that being adding to the end).

pkg/util/executor/remote.go
```diff
@@ -102,3 +102,7 @@ func fetchPod(ctx context.Context, client *kubernetes.Clientset, nodeName string
 
 	return &matchingPods[0], nil
 }
+
+func (r *RemoteCommandExecutor) CheckReady() ([]byte, error) {
+	return r.Run("/usr/bin/file", []string{"/tmp/ready"})
```
tserong, line 107: What's responsible for creating `/tmp/ready`, so that this will succeed?

tserong (approved):
I've read the HEP (https://github.com/harvester/harvester/pull/4833), and I've reviewed all the code (I went commit by commit because it's easier to try to understand that way), and this all looks well thought out, and like it does what it says. Also the tests are passing :-) But I have to disclaim that as I'm still relatively new to the codebase, there's things I don't really understand (e.g. exactly how kubevirt device plugins work), and as it's a big review there might be stuff I missed.

I've got a couple of small suggestions and a couple of questions/comments, but in general LGTM.

pkg/util/executor/remote.go
WebberHuang1118: I think it's just for checking the RemoteCommandExecutor, it doesn't need to exist?

tserong: My reading of this code is that it runs `/usr/bin/file /tmp/ready` inside a container. If `/tmp/ready` doesn't exist inside that container, that command will fail, and thus `CheckReady()` will return an error.

ibrokethecloud: yes. The driver-toolkits entrypoint writes this file after the GPU driver has been installer and kernel modules have been loaded. This was to ensure that command is not tried until everything is setup.

ibrokethecloud: sample script: https://github.com/ibrokethecloud/driver-packaging/blob/main/nvidia-driver/entrypoint.sh and will be moved to driver toolkit.

tests/snapshots/vgpu-node.umockdev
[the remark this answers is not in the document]
ibrokethecloud: only way to easily mock some tests locally and failing tests quicker.

pkg/util/executor/remote.go
tserong: Makes sense, thanks for the explanation.

pkg/controller/gpudevice/vgpu_controller.go
```diff
@@ -0,0 +1,357 @@
...
+
+	for _, v := range vGPUList {
+		parentDeviceEnabled, err := h.isParentGPUEnabled(v.Spec.ParentGPUDeviceAddress)
+		if err != nil {
+			return err
+		}
+
+		// if parentDevice is enabled, then skip deletion
+		// as controller will reconcile and re-configure vGPU when device parent SRIOVGPU device
+		// is reconfigured post reboot
```
WebberHuang1118, line 106: Can you help to explain how SRIOVGPU trigger the reconcile on VGPU? Through the node controller's [periodic reconciliation every 30 seconds](https://github.com/ibrokethecloud/pcidevices/blob/55de4d75a7cd9fac885bbad31306ff83e0808bd7/pkg/controller/nodes/node_controller.go#L123)?

pkg/controller/gpudevice/vgpu_controller.go
```diff
@@ -0,0 +1,357 @@
...
+	if err != nil {
+		return vgpuObj, err
+	}
+
+	return h.reconcileDisabledVGPUStatus(vgpuObj)
+}
+
+func (h *Handler) patchKubevirtCR() error {
+	return nil
+}
```
WebberHuang1118, line 192: This function seems not used.

pkg/deviceplugins/vgpu_device_manager.go
```diff
@@ -0,0 +1,499 @@
...
+		}
+	}
+	logrus.Debugf("found device count %d for plugin %s", count, dp.resourceName)
+	return count
+}
+
+func NewVGPUDevicePlugin(ctx context.Context, vGPUList []string, resourceName string) *VGPUDevicePlugin {
+	serverSock := SocketPath(strings.Replace(resourceName, "/", "-", -1))
+
+	initHandler()
```
WebberHuang1118, line 79: Can you help explain if initHandler() is needed? IIUC, this function seems not to take effect. Thanks.

pkg/deviceplugins/vgpu_device_manager.go
```diff
@@ -0,0 +1,499 @@
...
+func (dp *VGPUDevicePlugin) SetStarted(stop chan struct{}) {
+	c := dp.starter
+	c.stopChan = stop
+	c.started = true
+	logrus.Infof("Started DevicePlugin: %s", dp.resourceName)
+}
+
+func (dp *VGPUDevicePlugin) Started() bool {
+	return dp.starter.started
+}
```
WebberHuang1118, line 122: Should this require a lock as [gets initialized](https://github.com/ibrokethecloud/pcidevices/blob/55de4d75a7cd9fac885bbad31306ff83e0808bd7/pkg/deviceplugins/vgpu_device_manager.go#L425-L429)?

pkg/deviceplugins/vgpu_device_manager.go
```diff
@@ -0,0 +1,499 @@
...
+			logrus.Debugf("trying to allocate device for %s", devID)
+			devicePath := filepath.Join(v1beta1.MdevRoot, devID)
+			_, err := os.Stat(devicePath)
+			if err != nil {
+				logrus.Errorf("error allocating device %s: %v", devID, err)
+				continue
+			}
+			deviceSpecs = append(deviceSpecs, &pluginapi.DeviceSpec{
+				ContainerPath: vfioDevicePath,
+				HostPath:      vfioDevicePath,
```
WebberHuang1118, line 234: Can you help explain why these two values are set to `/dev/vfio`? Thanks

pkg/deviceplugins/vgpu_device_manager.go
```diff
@@ -0,0 +1,499 @@
...
+			} else if event.Name == dp.socketPath && event.Op == fsnotify.Remove {
+				logrus.Infof("device socket file for device %s was removed, kubelet probably restarted.", dp.resourceName)
+				return nil
+			}
+		}
+	}
+}
+func (dp *VGPUDevicePlugin) GetDevicePath() string {
+	return dp.devicePath
+}
```
WebberHuang1118, line 356: This function seems not used.

pkg/deviceplugins/vgpu_device_manager.go
```diff
@@ -0,0 +1,499 @@
...
+func (dp *VGPUDevicePlugin) PreStartContainer(_ context.Context, _ *pluginapi.PreStartContainerRequest) (*pluginapi.PreStartContainerResponse, error) {
+	res := &pluginapi.PreStartContainerResponse{}
+	return res, nil
+}
+
+func (dp *VGPUDevicePlugin) GetInitialized() bool {
+	dp.lock.Lock()
+	defer dp.lock.Unlock()
+	return dp.initialized
+}
```
WebberHuang1118, line 429: This function seems not used.

pkg/util/executor/remote.go
```diff
@@ -0,0 +1,108 @@
...
+			matchingPods = append(matchingPods, v)
+		}
+	}
+
+	if len(matchingPods) == 0 || len(matchingPods) > 1 {
+		return nil, fmt.Errorf("expected to find exactly one pod for nvidia driver on node %s, but got %d", nodeName, len(matchingPods))
+	}
+
+	return &matchingPods[0], nil
+}
```
WebberHuang1118, line 104: This function seems not used.

WebberHuang1118 (commented):
LGTM, just please help to answer some questions, thanks.

pkg/deviceplugins/vgpu_device_manager.go
```diff
@@ -0,0 +1,482 @@
+package deviceplugins
+
+/* This file was part of the KubeVirt project, copied to this project
+ * to get around private package issues.
+ *
```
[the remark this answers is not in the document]
ibrokethecloud: i am sure the issue was use of `internal` packages. I have now forgotten which package this was.

pkg/deviceplugins/vgpu_device_manager.go
ibrokethecloud: the lock is used in vgpu_controller.go at the time of createOrUpdateDevicePlugin: https://github.com/ibrokethecloud/pcidevices/blob/vgpu/pkg/controller/gpudevice/vgpu_controller.go#L241-L243

pkg/deviceplugins/vgpu_device_manager.go
```diff
@@ -0,0 +1,499 @@
...
+	if dp != nil {
+		logrus.Infof("Removing %s from device plugin", uuid)
+		dp.MarkVGPUDeviceAsUnHealthy(uuid)
+	}
+
+	for i, dev := range dp.devs {
+		if dev.ID == uuid {
+			dp.devs[i].Health = pluginapi.Unhealthy
+		}
+	}
```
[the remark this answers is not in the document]
ibrokethecloud: This was to ensure this is device is marked unhealthy before unbinding / removing vgpu. This is to just handle edge cases where reconcile may fail for any reason and avoid device from being used in scheduling decisions.

pkg/util/executor/interface.go
```diff
@@ -0,0 +1,39 @@
...
+	sriovManageCommand  = "/usr/lib/nvidia/sriov-manage"
+)
+
+func (l *LocalExecutor) Run(cmd string, args []string) ([]byte, error) {
+	// localExecutor is run inside pcidevices pod, so need to add `/host` to command
+	cmd = filepath.Join(localExecutorAppend, cmd)
+	localCommand := exec.Command(cmd, args...)
+	localCommand.Env = append(localCommand.Env, l.envVars...)
+	return localCommand.Output()
+}
```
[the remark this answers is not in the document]
ibrokethecloud: Originally the nvidia-drivers were installed on host using `after_install_chroot_commands` section but this presented issues during harvester upgrade etc as OS level changes would be lost.

```
scheme_version: 1
os:
  after_install_chroot_commands:
  - /usr/bin/curl -o /tmp/NVIDIA-Linux-x86_64-525.60.12-vgpu-kvm.run http://172.19.96.23/iso/vgpu/NVIDIA-Linux-x86_64-525.60.12-vgpu-kvm.run
  - /usr/bin/chmod +x /tmp/NVIDIA-Linux-x86_64-525.60.12-vgpu-kvm.run
  - /tmp/NVIDIA-Linux-x86_64-525.60.12-vgpu-kvm.run -q --ui=none
 ```

The only reason for leaving the implementation is to allow for future driver support / device support such as non nvidia gpus etc.

File: pkg/util/executor/remote.go (hunk 10 above)
Comment (ibrokethecloud) [reply pkg/util/executor/remote.go:104]: good point. I had moved the logic into the controller.

File: pkg/controller/gpudevice/gpu_controller.go [hunk 14]
```diff
@@ -0,0 +1,255 @@
...
+	_, err := h.sriovGPUCache.Get(gpu.Name)
+	if err != nil {
+		if apierrors.IsNotFound(err) {
+			_, createErr := h.sriovGPUClient.Create(gpu)
+			return createErr
+		}
+	}
+
+	return err
+}
```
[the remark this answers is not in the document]
ibrokethecloud: Initial plan was to reconcile the real world and k8s api state accordingly. reconciling from host back into k8s api in initial versions led to issues post reboot. During a reboot the sriov devices are removed and gpu is unconfigured. This would result in gpu getting updated and vgpu devices being removed since gpu was now unconfigured. As a result the logic was removed to ensure we reconcile only from defined k8s api state to the host.

pkg/deviceplugins/vgpu_device_manager.go
ibrokethecloud: they are both the same since we run the pcidevice plugin as a privileged container so it has direct access to the /dev tree on the host.

pkg/controller/gpudevice/vgpu_controller.go
ibrokethecloud: good spot. I had moved the logic to another method.

pkg/deviceplugins/vgpu_device_manager.go
ibrokethecloud: yeah not needed. legacy stuff that I have now cleaned up.

pkg/deviceplugins/vgpu_device_manager.go
ibrokethecloud: cleaned up.

pkg/controller/gpudevice/vgpu_controller.go
```diff
@@ -0,0 +1,357 @@
...
+	}
+	// Start the plugin
+	stop := make(chan struct{})
+	go func() {
+		err := dp.Start(stop)
+		if err != nil {
+			logrus.Errorf("error starting %s device plugin: %s", dp.GetDeviceName(), err)
+		}
+		// TODO: test if deleting this stops the DevicePlugin
+		<-stop
```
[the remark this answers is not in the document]
ibrokethecloud: The channel is to ensure plugin does not close until the controller has reconciled all devices and has ensured this is not needed. This actual stop is called when needed from here: https://github.com/ibrokethecloud/pcidevices/blob/vgpu/pkg/controller/gpudevice/vgpu_controller.go#L210

pkg/controller/gpudevice/vgpu_controller.go
ibrokethecloud: the node object is a fake crd with only one purpose which is to regularly reconcile hardware objects to detect changes in underlying OS. In some cases like sriovnetworkdevices and gpu's if you passthrough the device directly to a guest then we should ignore it from usage in srviogpudevices and subsequent vgpudevices of Network VF's as the device is no longer being managed the host. The reverse is also true, if the GPU or NIC is used for VF's then PCIDevice object for this should not exist to avoid the PF being passed through to a guest.

pkg/controller/gpudevice/vgpu_controller.go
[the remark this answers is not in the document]
ibrokethecloud: based on our discussion we should probably track and change the logic to close these subroutines.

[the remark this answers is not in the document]
WebberHuang1118: The unreclaimed goroutine seems no harm so far, I could check it in the laterly. Thanks.

pkg/deviceplugins/vgpu_device_manager.go
```diff
@@ -0,0 +1,488 @@
...
+	defer ticker.Stop()
+	select {
+	case <-dp.deregistered:
+	case <-ticker.C:
+	}
+
+	dp.server.Stop()
+	dp.setInitialized(false)
+	return dp.cleanup()
+}
```
WebberHuang1118, line 376: I think I found a possible deadlock here, which is hidden by the ticker. Pcidevice controller waits for the chan `dp.deregistered`, but it only be notified if `ListAndWatch()` is [knocked by chan done](https://github.com/ibrokethecloud/pcidevices/blob/3a762072ebdd1506de8db95f5b829ea05f48e23d/pkg/deviceplugins/vgpu_device_manager.go#L197-L198)  and [close chan `dp.deregistered`](https://github.com/ibrokethecloud/pcidevices/blob/3a762072ebdd1506de8db95f5b829ea05f48e23d/pkg/deviceplugins/vgpu_device_manager.go#L209), however, `stopDevicePlugin()` never close chan `dp.done` first.

As a result, `stopDevicePlugin()` relies on ticker timeout to stop the server and remove the socket file, after this `ListAndWatch()` finally gets the chance to [get out of the loop](https://github.com/ibrokethecloud/pcidevices/blob/3a762072ebdd1506de8db95f5b829ea05f48e23d/pkg/deviceplugins/vgpu_device_manager.go#L183-L203) and try to [send empty list response](https://github.com/ibrokethecloud/pcidevices/blob/3a762072ebdd1506de8db95f5b829ea05f48e23d/pkg/deviceplugins/vgpu_device_manager.go#L206-L208), but during this phase, the server already torn down, so it will receive err message `nvidia.com/NVIDIA_xx-xxx device plugin failed to deregister: rpc error: code = Unavailable desc = transport is closing`

ibrokethecloud: this should now be fixed with some minor tweaks to logic in `stopDevicePlugin`

WebberHuang1118 (approved):
LGTM, thanks.

bk201 (approved):
just need to fix ci

PR #63, opened 2024-01-05
feat: bump golangci-lint to v1.55.2

FrankYang0529:
**Problem:**
The golangci-lint version before v1.51.0 may have OOM issue with golang v1.20.

**Solution:**
Bump golangci-lint to a version after v1.51.0.

**Related Issue:**
https://github.com/harvester/harvester/issues/4938

### DIFF
--- a/Dockerfile.dapper
+++ b/Dockerfile.dapper
@@ -15,9 +15,7 @@ RUN export K8S_VERSION=1.24.2 && \
     tar -C /usr/local/kubebuilder --strip-components=1 -zvxf envtest-bins.tar.gz
 
 ## install golangci
-RUN if [ "${ARCH}" = "amd64" ]; then \
-    curl -sSfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh | sh -s v1.49.0; \
-    fi
+RUN curl -sSfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh | sh -s v1.55.2
 
 ENV GO111MODULE on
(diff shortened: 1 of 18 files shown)


ibrokethecloud (commented):
should we setup linter to only warn about `dot-imports`? At present if we start changing dot-imports there will be a lot of work rebasing PR's since just about every integration test leveraging Ginkgo/Gomega is using the `dot-import`.

ibrokethecloud (commented):
> should we setup linter to only warn about `dot-imports`? At present if we start changing dot-imports there will be a lot of work rebasing PR's since just about every integration test leveraging Ginkgo/Gomega is using the `dot-import`.  @bk201?

ibrokethecloud:
dropping a `revive.toml` in project root, with the following allows us to skip checks: https://github.com/mgechev/revive/blob/master/RULES_DESCRIPTIONS.md#dot-imports

```more revive.toml
[rule.dot-imports]
  arguments = [{ allowedPackages = ["github.com/onsi/ginkgo/v2","github.com/onsi/gomega"] }]
```

FrankYang0529:
> should we setup linter to only warn about `dot-imports`? At present if we start changing dot-imports there will be a lot of work rebasing PR's since just about every integration test leveraging Ginkgo/Gomega is using the `dot-import`.

I use another to solve `dot-imports`. May you take a look? Thank you.

ibrokethecloud (approved):
lgtm. thanks.

bk201 (approved):

PR #65, opened 2024-02-07
change logic to filter MIG GPUs

ibrokethecloud:

**Problem:**

NVIDIA GPU's supporting MIG mode vGPU's are also detected as SRIOVGPUDevices.

**Solution:**

PR introduces logic to filter out NVIDIA GPU's which support MIG mode vGPU's only.

Currently these GPU's also have a sriov_totalvfs file as a result of which the devices show up as `SRIOVGPUDevices`.

The check has been changed to check `sriov_vf_device` file which contains the vf device ID. 

On MIG devices this contains `0` while on GPU's supporting vGPU's this contains a hexadecimal device id.

**Related Issue:**
https://github.com/harvester/harvester/issues/5112
**Test plan:**

### DIFF
--- a/pkg/util/common/common.go
+++ b/pkg/util/common/common.go
@@ -9,13 +9,14 @@ import (
 )
 
 const (
-	defaultTotalVFFile      = "sriov_totalvfs"
 	defaultConfiguredVFFile = "sriov_numvfs"
+	defaultVFCheckFile      = "sriov_vf_device"
 )
 
-// IsDeviceSRIOVCapable checks for existence of `sriov_totalvfs` file in the pcidevice tree
+// IsDeviceSRIOVCapable checks for existence of `sriov_vf_device` file in the pcidevice tree
 func IsDeviceSRIOVCapable(devicePath string) (bool, error) {
-	_, err := os.Stat(filepath.Join(devicePath, defaultTotalVFFile))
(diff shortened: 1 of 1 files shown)


pkg/util/common/common.go
```diff
@@ -9,20 +9,29 @@ import (
 )
 
 const (
-	defaultTotalVFFile      = "sriov_totalvfs"
 	defaultConfiguredVFFile = "sriov_numvfs"
+	defaultVFCheckFile      = "sriov_vf_device"
 )
 
 // IsDeviceSRIOVCapable checks for existence of `sriov_totalvfs` file in the pcidevice tree
```
WebberHuang1118, line 16: Maybe change the comment accordingly

WebberHuang1118 (approved):
LGTM, just a nit, thanks.

pkg/util/common/common.go
WebberHuang1118, line 16: Maybe change the comment accordingly

WebberHuang1118:
https://github.com/ibrokethecloud/pcidevices/blob/0713f79b972b1fa3d54613a577fc8f113a6b7c46/pkg/util/common/common.go#L16
Maybe change the comment accordingly

tserong (approved):

PR #66, opened 2024-02-19
Fixing pcidevice device plugin stop deadlock

WebberHuang1118:
**Problem:**
pci controller stops the device plugin with potential deadlock

**Solution:**
Refactor the device plugin stop flow to terminate it gracefully, the flow would like this:
- Device plugin [close chan dp.done](https://github.com/harvester/pcidevices/blob/a42eb589df13728a1e91a374d6734a70aedf3d8f/pkg/deviceplugins/device_manager.go#L398-L400) and [wait the notification from chan dp.deregistered](https://github.com/harvester/pcidevices/blob/a42eb589df13728a1e91a374d6734a70aedf3d8f/pkg/deviceplugins/device_manager.go#L406)
- `healthCheck()\performCheck()` is [kicked by chan dp.done and gets out of the loop](https://github.com/harvester/pcidevices/blob/a42eb589df13728a1e91a374d6734a70aedf3d8f/pkg/deviceplugins/device_manager.go#L359-L360)
- `ListAndWatch()` is [kicked by chan dp.done and gets out of the loop](https://github.com/harvester/pcidevices/blob/a42eb589df13728a1e91a374d6734a70aedf3d8f/pkg/deviceplugins/device_manager.go#L239-L240)
- `ListAndWatch()` [sends an empty list response and closes chan dp.deregistered](https://github.com/harvester/pcidevices/blob/a42eb589df13728a1e91a374d6734a70aedf3d8f/pkg/deviceplugins/device_manager.go#L250-L253)
- `stopDevicePlugin()` is [notified by chan dp.deregistered](https://github.com/harvester/pcidevices/blob/a42eb589df13728a1e91a374d6734a70aedf3d8f/pkg/deviceplugins/device_manager.go#L406) and [stops the server gracefully](https://github.com/harvester/pcidevices/blob/a42eb589df13728a1e91a374d6734a70aedf3d8f/pkg/deviceplugins/device_manager.go#L410-L412)

**Related Issue:**
https://github.com/harvester/harvester/issues/5164

**Test Plan:**
- Enable and disable one pcidevice passthrough
- There should be not error message similar as `device plugin failed to deregister: rpc error: code = Unavaila  
│ ble desc = transport is closing","pos":"device_manager.go:250","reason":"rpc error: code = Unavailable desc = transport is closing"` in the pci controller daemonset

### DIFF
--- a/pkg/controller/nodes/node_controller.go
+++ b/pkg/controller/nodes/node_controller.go
@@ -7,8 +7,6 @@ import (
 	"reflect"
 	"time"
 
-	"github.com/harvester/pcidevices/pkg/controller/gpudevice"
-
 	ctlnetworkv1beta1 "github.com/harvester/harvester-network-controller/pkg/generated/controllers/network.harvesterhci.io/v1beta1"
 	"github.com/jaypipes/ghw"
 	ctlcorev1 "github.com/rancher/wrangler/pkg/generated/controllers/core/v1"
@@ -17,6 +15,7 @@ import (
 	"k8s.io/apimachinery/pkg/labels"
 
 	"github.com/harvester/pcidevices/pkg/apis/devices.harvesterhci.io/v1beta1"
+	"github.com/harvester/pcidevices/pkg/controller/gpudevice"
 	"github.com/harvester/pcidevices/pkg/controller/pcidevice"
 	"github.com/harvester/pcidevices/pkg/controller/sriovdevice"
 	ctl "github.com/harvester/pcidevices/pkg/generated/controllers/devices.harvesterhci.io/v1beta1"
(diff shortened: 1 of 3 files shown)


Yu-Jack:
I think we just need to add following codes before L166 without deleting all `stop` channel.

```go
if !dp.starter.started {
    return nil
}

close(dp.starter.stopChan)
dp.starter.started = false
```

https://github.com/harvester/pcidevices/blob/feb7e8ce680c53e5ca0e9925179938ebb24ecb72/pkg/deviceplugins/device_manager.go#L164-L166

The reason is `stop` and `done` are kind of different thoughts here.
- `stop` is more like someone controls this object from outside, it will tell this object should be stopped.
- `done` is more like object inner status, it tells whole object about current progress.

Just like I mentioned in that PR https://github.com/harvester/pcidevices/pull/67#issuecomment-1970802345, our structure lacks of some objects, but we still used the similar flow of KubeVirt's device plugin. I think we could keep it original flow for now until we have time to redesign it based on our structure.

Yu-Jack:
BTW, this is commit what I changed (https://github.com/Yu-Jack/pcidevices/commit/bac0ffd250d9e957c141f4001086b4d05751d9a2)

bk201:
The parent issue is closed; we can revisit this if needed. cc @Vicente-Cheng

PR #67, opened 2024-02-20
Pci device plugin fix orphan goroutine and remove unused channel

WebberHuang1118:
**Problem:**
After a PCI device plugin is disabled, there is Go routine (invoked as the device plugin started) [blocked by chan stop](https://github.com/harvester/pcidevices/blob/82535bf09278420d225eaf3fe8dbff2e450b8f71/pkg/controller/pcideviceclaim/pcideviceclaim_controller.go#L429-L436), however, chan stop is never touched, this brings an orphan Go routine once a PCI device is enabled and disabled.

**Solution:**
PCI device plugin fix the orphan goroutine and removes unused channel

**Related Issue:**
https://github.com/harvester/harvester/issues/5179

**Additional Context:**
The CodeFactor `Complex Method` error could be fixed once PR https://github.com/harvester/pcidevices/pull/66 merged

### DIFF
--- a/Dockerfile.dapper
+++ b/Dockerfile.dapper
@@ -9,10 +9,8 @@ RUN zypper -n ar  https://download.opensuse.org/repositories/hardware/15.6/hardw
 RUN go install golang.org/x/lint/golint@latest
 RUN go install golang.org/x/tools/cmd/goimports@latest
 RUN go install github.com/incu6us/goimports-reviser/v3@v3.10.0
-RUN export K8S_VERSION=1.24.2 && \
-    curl -sSLo envtest-bins.tar.gz "https://go.kubebuilder.io/test-tools/${K8S_VERSION}/$(go env GOOS)/$(go env GOARCH)" && \
-    mkdir /usr/local/kubebuilder && \
-    tar -C /usr/local/kubebuilder --strip-components=1 -zvxf envtest-bins.tar.gz
+RUN go install sigs.k8s.io/controller-runtime/tools/setup-envtest@latest && \
+    setup-envtest use 1.31.x --bin-dir /usr/local/kubebuilder -p path
 
 ## install golangci
 RUN curl -sSfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh | sh -s latest
--- a/pkg/controller/gpudevice/vgpu_controller.go
+++ b/pkg/controller/gpudevice/vgpu_controller.go
@@ -8,14 +8,12 @@ import (
 	"reflect"
 	"sync"
 
+	"github.com/google/uuid"
+	"github.com/sirupsen/logrus"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
-
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
 	kubevirtv1 "kubevirt.io/api/core/v1"
-
-	"github.com/google/uuid"
-	"github.com/sirupsen/logrus"
 
 	"github.com/harvester/pcidevices/pkg/apis/devices.harvesterhci.io/v1beta1"
 	"github.com/harvester/pcidevices/pkg/deviceplugins"
--- a/pkg/controller/pcideviceclaim/pcideviceclaim_controller.go
+++ b/pkg/controller/pcideviceclaim/pcideviceclaim_controller.go
@@ -491,17 +491,14 @@ func (h *Handler) startDevicePlugin(
 	if dp.Started() {
 		return nil
 	}
-	// Start the plugin
-	stop := make(chan struct{})
+
 	go func() {
-		err := dp.Start(stop)
+		err := dp.Start()
 		if err != nil {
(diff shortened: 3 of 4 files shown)


Yu-Jack:
I traced the original `stop` usage in kubevirt.

The [heartBeat](https://github.com/kubevirt/kubevirt/blob/2bb88c3d35d33177ea16c0f1e9fffdef1fd350c6/pkg/virt-handler/heartbeat/heartbeat.go#L122) will trigger kubevirt to [refresh the permitted devices](https://github.com/kubevirt/kubevirt/blob/2bb88c3d35d33177ea16c0f1e9fffdef1fd350c6/pkg/virt-handler/device-manager/device_controller.go#L241-L247).
-> Then it will [start and stop all devices](https://github.com/kubevirt/kubevirt/blob/2bb88c3d35d33177ea16c0f1e9fffdef1fd350c6/pkg/virt-handler/device-manager/device_controller.go#L281-L290).
-> Then the `DeviceController` uses [`controlledDevice` to close stopChan](https://github.com/kubevirt/kubevirt/blob/2bb88c3d35d33177ea16c0f1e9fffdef1fd350c6/pkg/virt-handler/device-manager/device_controller.go#L89-L97)

In device_plugin side, you could see it [uses `errChan` to wait](https://github.com/kubevirt/kubevirt/blob/2bb88c3d35d33177ea16c0f1e9fffdef1fd350c6/pkg/virt-handler/device-manager/pci_device.go#L114)
-> and there is a [healthCheck function to detect stopChan](https://github.com/kubevirt/kubevirt/blob/2bb88c3d35d33177ea16c0f1e9fffdef1fd350c6/pkg/virt-handler/device-manager/pci_device.go#L225) to return nil if stopChan is closed.

The structure is one `DeviceController` uses multiple `controlledDevice`, and one `controlledDevice` is mapped to one `devicePlugin`. But, in our structure, we just used `devicePlugin`, I think that's why `stop` chan is weird in our structure.

WebberHuang1118:
> I traced the original `stop` usage in kubevirt.
> 
> The [heartBeat](https://github.com/kubevirt/kubevirt/blob/2bb88c3d35d33177ea16c0f1e9fffdef1fd350c6/pkg/virt-handler/heartbeat/heartbeat.go#L122) will trigger kubevirt to [refresh the permitted devices](https://github.com/kubevirt/kubevirt/blob/2bb88c3d35d33177ea16c0f1e9fffdef1fd350c6/pkg/virt-handler/device-manager/device_controller.go#L241-L247). -> Then it will [start and stop all devices](https://github.com/kubevirt/kubevirt/blob/2bb88c3d35d33177ea16c0f1e9fffdef1fd350c6/pkg/virt-handler/device-manager/device_controller.go#L281-L290). -> Then the `DeviceController` uses [`controlledDevice` to close stopChan](https://github.com/kubevirt/kubevirt/blob/2bb88c3d35d33177ea16c0f1e9fffdef1fd350c6/pkg/virt-handler/device-manager/device_controller.go#L89-L97)
> 
> In device_plugin part, it [uses `errChan` to wait](https://github.com/kubevirt/kubevirt/blob/2bb88c3d35d33177ea16c0f1e9fffdef1fd350c6/pkg/virt-handler/device-manager/pci_device.go#L114) which waits for grpc server and health check error in Start function. -> and there is a [healthCheck function to detect stopChan](https://github.com/kubevirt/kubevirt/blob/2bb88c3d35d33177ea16c0f1e9fffdef1fd350c6/pkg/virt-handler/device-manager/pci_device.go#L225) to return nil if stopChan is closed.
> 
> The structure is one `DeviceController` uses multiple `controlledDevice`, and one `controlledDevice` is mapped to one `devicePlugin`. But, in our structure, we just used `devicePlugin`, and our device plugin [Start function doesn't wait for health check](https://github.com/harvester/pcidevices/blob/feb7e8ce680c53e5ca0e9925179938ebb24ecb72/pkg/deviceplugins/device_manager.go#L169), so I think that's why `stop` chan is weird in our structure.

@Yu-Jack, Thanks for your explanation,

IIUC, in the current implementation `stop chan` doesn't take effect, its existence is just because the PCI controller references some Kubevirt code snippet from the beginning.

IMHO, since this PR doesn't introduce any structural or control flow change, I'd like to remove the unused code path. WDYT? @bk201 @ibrokethecloud , thanks.

Yu-Jack:
Just leave link to original comment https://github.com/harvester/pcidevices/pull/66#issuecomment-1972737652 if we don't want to remove existing codes.

Yu-Jack (commented):
I prefer not to delete stop channel. This stop is a control flag to control device plugin. For example, if we remove this stop, it can't return in health check function. Instead, we can remove done channel in this case.

ibrokethecloud:
Looks like one channel is enough. we could either use done or stop to trigger cleanup. that channel should be used in the healthcheck.

this is legacy stuff which we copied over when we tried to setup the device plugins.

Thanks for the PR @WebberHuang1118 to help clean this up.

bk201:
The parent issue is closed; we can revisit this if needed. cc @Vicente-Cheng @pohanhuang

PR #75, opened 2024-04-22
migrate from drone ci to gh actions

ibrokethecloud:

**Problem:**

PR migrations CI from drone to gh actions
**Solution:**

**Related Issue:**
https://github.com/harvester/harvester/issues/5327
**Test plan:**


scripts/validate
```diff
@@ -15,5 +15,5 @@ if [[ -z "$(command -v golangci-lint)" ]]; then
 fi
 
 echo "Running: golangci-lint run"
-golangci-lint run
+golangci-lint run --timeout 10m
```
WebberHuang1118, line 18: Hi @ibrokethecloud,
Can you help to explain why we need to introduce this timeout? Thanks.

ibrokethecloud: the default timeout is 1m, and I noticed random failures due to golangci-lint timeout. This is likely due to the lack of resources on the runners.

Yu-Jack (commented):
Same as others, we should generate amd64 and arm64 build, and move specified binary in Dockerfile.

.github/workflows/template-build.yml
```diff
@@ -0,0 +1,58 @@
...
+        with:
+          provenance: false
+          context: .
+          platforms: linux/amd64
+          load: true
+          file: package/Dockerfile
+          push: ${{ inputs.push }}
+          tags: ${{env.imageName}}:${{ inputs.release-tag-name }}
+          build-args: |
+            VERSION=${{ github.ref_name }}-${{ github.sha }}-head
```
Yu-Jack, line 58: BTW, this one isn't used in Dockerfile as well, so we could remove it.

ibrokethecloud: 👍

Yu-Jack (approved):
LGTM, thanks.

WebberHuang1118 (approved):
LGTM, thanks.

PR #78, opened 2024-05-22
Feat: add USB Passthrough

Yu-Jack:
**Problem:**

This is a new feature request to make use of [KubeVirt USB Passthrough on the node](https://kubevirt.io/user-guide/virtual_machines/host-devices/#usb-host-passthrough).

**Solution:**

Expose KubeVirt config and create custom external device plugin to select specified usb device.

**Related Issue:**

https://github.com/harvester/harvester/issues/5762

**Guideline to this PR**

This PR is quite large, so I list some files as entry for reviewing.

- New APIs
  - pkg/apis/devices.harvesterhci.io/v1beta1/usb_device.go
  - pkg/apis/devices.harvesterhci.io/v1beta1/usb_device_claim.go
- Controllers
  - Could check test cases first.
  - pkg/controller/usbdevice/usbdevice_controller.go
  - pkg/controller/usbdevice/usbdevice_claim_controller.go
- Custom USB Device Plugin
  - pkg/deviceplugins/usb_device_plugin.go
- Retrive USB detail informations
  - pkg/util/gousb/usbid/load.go
- Webhook
  - pkg/webhook/usbdevice.go
  - pkg/webhook/usbdeviceclaim.go
  - pkg/webhook/vmi_validatation.go
  - pkg/webhook/pcideviceclaim.go

Other files are targeted to different purposes, such as rebooting,  code generator and fixing codegen.

This is a very simple arch of objects. 

image

**Test plan:**

Please find machine with USB device on the node to test this feature.

This is happy path: 
1.  There should be devices on the `usbdevice` custom resource.
2. Enable it by creating like following custom resource
```yaml
apiVersion: devices.harvesterhci.io/v1beta1
kind: USBDeviceClaim
metadata:
  name: {usbdevice.Name}
  ownerReferences:
  - apiVersion: devices.harvesterhci.io/v1beta1
    kind: USBDevice
    name: {usbdevice.Name}
    uid: {usbdevice.UID}
```
3. Create a VM, and stop it. Then modify its yaml.
```
domain:
  device:
    hostDevices:
      - deviceName: kubevirt.io/{usbdevice.Name}
         name: {usbdevice.Name}
```
4. Start vm, and run `lsusb` to check device.

The others paths, including:
1. Two same vendor and product, should be able pick one of them.
2. Delete the `usbdeviceclaim` which is being used, should not be deleted.
3. Delete the `usbdevice` which is being used, should not be deleted.
4. Enable `pcidevice` after its `usbdevice` is claimed, should not be enabled.


pkg/crd/crd.go
```diff
@@ -126,6 +126,23 @@ func List() []crd.CRD {
...
+				WithColumn("Description", ".status.description").
+				WithColumn("Resource Name", ".status.resourceName").
+				WithColumn("PCI Address", ".status.pciAddress").
+				WithColumn("Enabled", ".status.enabled")
+		}),
+		newCRD(&devices.USBDeviceClaim{}, func(c crd.CRD) crd.CRD {
+			c.NonNamespace = true
+			return c.
+				WithColumn("Node Name", ".status.nodeName").
+				WithColumn("PCI Address", ".status.pciAddress")
```
ibrokethecloud, line 144: should we also include `ClaimedBy` here? or change field to `UserName` to be more in sync with PCIDeviceClaims?

pkg/controller/usbdevice/usbdevice_controller.go
```diff
@@ -0,0 +1,217 @@
...
+	}
+
+	if ud.Status.NodeName == cl.nodeName {
+		udcList, err := h.usbClaimClient.List(metav1.ListOptions{LabelSelector: cl.selector()})
+		if err != nil {
+			logrus.Errorf("error listing USBDeviceClaims during device watch: %v", err)
+			return nil, err
+		}
+
+		var rr []relatedresource.Key
```
ibrokethecloud, line 68: we are requeueing all usbdeviceclaims. should we not filter on only the claim where pcidevice address matches the usb device pcidevice address? we could even label the usbdeviceclaim with pcideviceaddress string to make it easy to filter using label selectors.

pkg/controller/usbdevice/usbdevice_controller.go
```diff
@@ -0,0 +1,217 @@
...
+	"k8s.io/apimachinery/pkg/runtime"
+
+	"github.com/harvester/pcidevices/pkg/apis/devices.harvesterhci.io/v1beta1"
+	"github.com/harvester/pcidevices/pkg/deviceplugins"
+	ctldevicerv1vbeta1 "github.com/harvester/pcidevices/pkg/generated/controllers/devices.harvesterhci.io/v1beta1"
+	"github.com/harvester/pcidevices/pkg/util/gousb"
+	"github.com/harvester/pcidevices/pkg/util/gousb/usbid"
+)
+
+type Handler struct {
```
ibrokethecloud, line 19: we should pass the cache interfaces for both types, and change all get/lists to use the cache

pkg/controller/usbdevice/usbdevice_claim_controller.go
```diff
@@ -0,0 +1,283 @@
...
+		devicePlugin:          map[string]*deviceController{},
+		devicePluginConvertor: devicePluginHelper,
+	}
+}
+
+func (h *ClaimHandler) OnUSBDeviceClaimChanged(_ string, usbDeviceClaim *v1beta1.USBDeviceClaim) (*v1beta1.USBDeviceClaim, error) {
+	if usbDeviceClaim == nil {
+		return usbDeviceClaim, nil
+	}
+
```
ibrokethecloud, line 63: we should add a check to ignore devices with a deletion timestamp

pkg/controller/usbdevice/usbdevice_claim_controller.go
```diff
@@ -0,0 +1,283 @@
...
+	return pluginDevice
+}
+
+func (h *ClaimHandler) OnRemove(_ string, claim *v1beta1.USBDeviceClaim) (*v1beta1.USBDeviceClaim, error) {
+	if claim == nil {
+		return claim, nil
+	}
+
```
ibrokethecloud, line 180: we need a check to ignore claims without a deletion timestamp

pkg/deviceplugins/usb_device_plugin_helper.go
```diff
@@ -0,0 +1,207 @@
...
+func DiscoverAllowedUSBDevices(usbs []v1.USBHostDevice) map[string][]*PluginDevices {
+	// The return value: USB USBDevice Plugins found and permitted to be exposed
+	plugins := make(map[string][]*PluginDevices)
+	// All USB devices found plugged in the Node
+	localDevices := discoverLocalUSBDevicesFunc()
+	for _, usbConfig := range usbs {
+		resourceName := usbConfig.ResourceName
+		// only accept ExternalResourceProvider: true for USB devices
+		if !usbConfig.ExternalResourceProvider {
+			logrus.Errorf("Skipping discovery of %s. To be handled by external device-plugin",
```
ibrokethecloud, line 192: in this case the device is being managed by the internal kubevirt device plugin, we should probably tweak the message to reflect the same 
```suggestion
			logrus.Errorf("Skipping discovery of %s. To be handled by kubevirt internally",
```

pkg/crd/crd.go
Yu-Jack: `UserName` is better!

pkg/controller/usbdevice/usbdevice_claim_controller.go
Yu-Jack: it's fixed, but didn't show out date .. weird

pkg/apis/devices.harvesterhci.io/v1beta1/usb_device_claim.go
```diff
@@ -0,0 +1,29 @@
...
+// +genclient
+// +genclient:nonNamespaced
+// +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
+// +kubebuilder:object:root=true
+
+type USBDeviceClaim struct {
+	metav1.TypeMeta   `json:",inline"`
+	metav1.ObjectMeta `json:"metadata,omitempty"`
+
+	Spec   USBDeviceClaimSpec   `json:"spec,omitempty"`
```
ibrokethecloud, line 14: `USBDeviceClaimSpec` serves no real purpose here. should we just drop it like the `usbdevice` crd?

Yu-Jack: yeah, need to do that.

pkg/deviceplugins/usb_device_plugin_helper.go
```diff
@@ -0,0 +1,208 @@
...
+		}
+	}
+
+	return &u
+}
+
+func parseSysUeventKeyValue(key string, value string, u *USBDevice) bool {
+	switch key {
+	case "BUSNUM":
+		val, err := strconv.ParseInt(value, 10, 32)
```
ibrokethecloud, line 53: we are skipping return usbdevice objects if there is an error for any reason. It makes sense to ignore the error but should we log the error atleast to help debug in case of edge cases?

Yu-Jack: sure, make sense, i'll add log

pkg/deviceplugins/usb_device_plugin.go
```diff
@@ -0,0 +1,469 @@
...
+	"kubevirt.io/kubevirt/pkg/safepath"
+	"kubevirt.io/kubevirt/pkg/util"
+	pluginapi "kubevirt.io/kubevirt/pkg/virt-handler/device-manager/deviceplugin/v1beta1"
+)
+
+var (
+	pathToUSBDevices = "/sys/bus/usb/devices"
+)
+
+var discoverLocalUSBDevicesFunc = discoverPluggedUSBDevices
```
ibrokethecloud, line 49: this rename is confusing, and makes it harder to follow? is there a specific reason for this?

Yu-Jack: Oh, it's for testing when doing mock in original kubevirt project. But, i didn't copy the test file here.

ibrokethecloud (approved):
lgtm. thanks.

pkg/deviceplugins/usb_device_plugin.go
```diff
@@ -0,0 +1,452 @@
...
+		}
+	}()
+
+	// Give the device plugin one second to properly deregister
+	ticker := time.NewTicker(1 * time.Second)
+	defer ticker.Stop()
+	select {
+	case <-plugin.deregistered:
+	case <-ticker.C:
+	}
```
WebberHuang1118, line 140: Just wait one second for the notification from `ListAndWatch()`, wouldn't it be too short?

WebberHuang1118 (commented):
General LGTM, just a few questions, thanks.

pkg/controller/usbdevice/usbdevice_controller.go
```diff
@@ -0,0 +1,266 @@
...
+)
+
+type DevHandler struct {
+	usbClient      ctldevicerv1vbeta1.USBDeviceClient
+	usbClaimClient ctldevicerv1vbeta1.USBDeviceClaimClient
+	usbCache       ctldevicerv1vbeta1.USBDeviceCache
+	usbClaimCache  ctldevicerv1vbeta1.USBDeviceClaimCache
+}
+
+var walkUSBDevices = deviceplugins.WalkUSBDevices
```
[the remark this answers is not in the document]
Yu-Jack: Yes. IIRC, this is called monkey patch. It's useful to do mock.

pkg/controller/usbdevice/usbdevice_claim_controller.go
```diff
@@ -0,0 +1,283 @@
...
+	"github.com/harvester/pcidevices/pkg/deviceplugins"
+	ctldevicerv1beta1 "github.com/harvester/pcidevices/pkg/generated/controllers/devices.harvesterhci.io/v1beta1"
+	ctlkubevirtv1 "github.com/harvester/pcidevices/pkg/generated/controllers/kubevirt.io/v1"
+)
+
+type DevClaimHandler struct {
+	usbClaimClient           ctldevicerv1beta1.USBDeviceClaimClient
+	usbClient                ctldevicerv1beta1.USBDeviceClient
+	virtClient               ctlkubevirtv1.KubeVirtClient
+	lock                     *sync.Mutex
```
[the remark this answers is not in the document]
Yu-Jack: In my opinion, I prefer not to use embedded struct in this case. I think it's more readable to keep it explicit.

pkg/controller/usbdevice/usbdevice_controller.go
```diff
@@ -0,0 +1,266 @@
...
+		if err != nil {
+			logrus.Errorf("failed to create USB device: %v\n", err)
+			return err
+		}
+
+		newOne.Status = usbDevice.Status
+		if _, err = h.usbClient.UpdateStatus(newOne); err != nil {
+			logrus.Errorf("failed to update new created USB device status: %v\n", err)
+			return err
+		}
```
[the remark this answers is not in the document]
Yu-Jack: Cause I tried to create a CR with up-to-date status before, but it failed. Did I miss something?

pkg/deviceplugins/usb_device_plugin.go
```diff
@@ -0,0 +1,452 @@
...
+
+func (plugin *USBDevicePlugin) StopDevicePlugin() error {
+	defer func() {
+		select {
+		case <-plugin.done:
+			return
+		default:
+			close(plugin.done)
+		}
+	}()
```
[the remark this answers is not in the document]
Yu-Jack: Seem this one is really not used ... interesting ... Let me check if we can remove it or not.

pkg/controller/usbdevice/usbdevice_claim_controller.go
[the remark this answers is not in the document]
WebberHuang1118: But would it be a kind of duplicate with `h.lock.Lock()` and `h.lock.Unlock()`? However, I'm okay with the current status. Just have your own choice, thanks.

pkg/controller/usbdevice/usbdevice_controller.go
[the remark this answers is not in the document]
WebberHuang1118: Resource update conflict?

pkg/deviceplugins/usb_device_plugin.go
WebberHuang1118: VGPU uses 5 second
https://github.com/Yu-Jack/pcidevices/blob/f22589e6ad959cad0d70987395c747b6833c7c5b/pkg/deviceplugins/vgpu_device_manager.go#L365-L370

@ibrokethecloud WDYT?

Yu-Jack: After discuss with @ibrokethecloud , we all agree with 5 seconds.

WebberHuang1118 (approved):
LGTM, thanks. Please check if need to squash commits.

PR #85, opened 2024-07-26
README: add more limitation about usb device [CI SKIP]

Yu-Jack:
Just revise some words.

### DIFF
--- a/README.md
+++ b/README.md
@@ -197,9 +197,11 @@ This controller primarily manages the lifecycle of the device plugin, ensuring t
 
 ## USB Device
 
-- Don't support live migration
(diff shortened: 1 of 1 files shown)


WebberHuang1118 (approved):
LGTM, thanks.

Yu-Jack:
I find out that rebooting **sometime** makes this failed, it's **not always**. 

For example, if usb device is `/002/002` at the first time, it won't change after rebooting. But, if we already re-plugged in the usb device **one time** to lead the path change (it might be `002/003` at that moment), then rebooting will reset the path to `/002/002`.

ibrokethecloud (approved):
lgtm. thanks.

PR #86, opened 2024-07-31
fix(usbdevice): close closed channel exception

Yu-Jack:
**Problem:**

Reproduce Steps

1. Plug in USB device A.
2. Unplug USB device A and plug it in again. This increases the device number.
3. Create a `usbdeviceclaim`.
4. Reboot the host.
5. Observe that the device number of USB device A is reset to the value from step 1.
6. Delete the `usbdeviceclaim`.
7. Observe the error from pcidevices-controller: close closed channel.

In this cause, the original `usbdeviceclaim` won't work due to device number changed. That's expected.

~~However, the `plugin.deregistered` channel is closed twice due to multiple calls to ListAndWatch. But, I cannot find the root cause yet.~~ Check https://github.com/harvester/pcidevices/pull/86#issuecomment-2275257248

https://github.com/user-attachments/assets/15f022ae-a313-498b-90eb-16a735cf1d59

**Solution:**

Do not use `close`, just make it buffered channel.

**Related Issue:**

https://github.com/harvester/harvester/issues/5762

**Test plan:**

Follow above steps, it shouldn't complain the same error.

https://github.com/user-attachments/assets/ab693cf2-b273-4031-a9a0-1e8462706152

### DIFF
--- a/pkg/deviceplugins/usb_device_plugin.go
+++ b/pkg/deviceplugins/usb_device_plugin.go
@@ -55,6 +55,7 @@ type USBDevicePlugin struct {
 	update       chan struct{}
 	deregistered chan struct{}
 	server       *grpc.Server
+	serverDone   chan struct{}
 	resourceName string
 	device       *PluginDevice
 	logger       *log.FilteredLogger
@@ -101,6 +102,10 @@ func (plugin *USBDevicePlugin) DeviceName() string {
 }
 
 func (plugin *USBDevicePlugin) stopDevicePlugin() error {
+	defer func() {
+		close(plugin.serverDone)
+	}()
+
 	// Give the device plugin one second to properly deregister
 	ticker := time.NewTicker(5 * time.Second)
 	defer ticker.Stop()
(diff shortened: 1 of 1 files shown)


ibrokethecloud (approved):

WebberHuang1118 (approved):
LGTM, thanks, do we need a separate issue for tracking the root cause?

Yu-Jack:
@ibrokethecloud @WebberHuang1118 

I find the root cause, it's quite straightforward.

We're trying to open gRPC server, and close it after error happens. Each time we start the gRPC server, kubelete invokes `ListAndWatch` function.

But, the program is stuck here since it didn't receive any value from that two channels. 

```go
done := false
for !done {
	select {
	case <-plugin.update:
		if err := sendUpdate(plugin.devicesToKubeVirtDevicePlugin()); err != nil {
			return err
		}
	case <-plugin.stop:
		done = true
}
```

So, I'll add new channel which presents that the gRPC server is closed or not. Please check my latest commits to see the changes, thanks!

WebberHuang1118 (approved):
LGTM, thanks.

ibrokethecloud (approved):
lgtm. thanks.

PR #89, opened 2024-09-09
fix: userName should be in spec instead of status

Yu-Jack:
**Problem:**
`UserName` should be in spec instead of status.

**Related Issue:**
https://github.com/harvester/dashboard/pull/1069#event-1417458412


bk201 (approved):

ibrokethecloud (approved):
lgtm. thanks.

PR #90, opened 2024-09-11
feat: add validation to check hostDevices[].name and .deviceName

Yu-Jack:
**Problem:**
Webhook should block that hostDevices[].name and  hostDevices[].deviceName is not matched.

**Related Issue:**
https://github.com/harvester/dashboard/pull/1069


Yu-Jack:
close it for figuring out more details

PR #91, opened 2024-09-13
initial work to reconcile devices from launcher pods

ibrokethecloud:

**Problem:**

Rancher allows provisioning of downstream clusters to leverage vGPUs.

When used with a machine deployment of more than 1 node, the actual gpu allocated is different from name specified since kubevirt leverages the DeviceName field to calculate launcher pod resource requirements, which are then subsequently used by schedular to identify correct nodes.

Since the actual name is not really used for device allocation, any random string can be used and this makes it difficult to track vGPU allocation in the cluster.

**Solution:**

PR introduces a minor change to leverage pod environment variables being set by the device plugin during the ContainerAllocateResponse. The additional vmi controller execs into the launcher pod to identify the ID set for resource, and subsequently map it to the GPU/Hostdevice resources. Once the devices are identify an annotation is set on the VM to provide info about actual device from a pool of devices passed through to the VM

For example for a host device this looks as follows:
```
harvesterhci.io/deviceAllocationDetails: '{"hostdevices":{"intel.com/8d26":["vgpu-01-0000001d0"]}}'
```

For GPU devices ithis looks as follows:
```
  harvesterhci.io/deviceAllocationDetails: '{"gpus":{"nvidia.com/NVIDIA_A2-2Q":["vgpu-01-000008005"]}}'
```

When a VM is shutdown the annotation is used to replace `name` for vGPU or host devices with actual device names from the deviceAllocationDetails annotation. This ensures that the VM can be edited and devices can be removed from Harvester UI post provisioning.

**Related Issue:**
https://github.com/harvester/harvester/issues/5774

**Test plan:**


ibrokethecloud:
backend changes related to: https://github.com/rancher/dashboard/pull/11399

pkg/controller/virtualmachine/virtualmachine.go
```diff
@@ -0,0 +1,475 @@
...
+	val, ok := vmObj.Annotations[v1beta1.DeviceAllocationKey]
+	if !ok {
+		// no device allocation annotations, nothing to do
+		logrus.WithFields(logrus.Fields{
+			"name":      vmi.Name,
+			"namespace": vmi.Namespace,
+			"hostname":  vmi.Status.NodeName,
+		}).Debug("skipping vmi as it has no device allocation annotation")
+		return vmi, nil
+	}
```
Yu-Jack, line 423: NIT: We could remove `if vmObj.Annotations == nil {` this condition to reduce some codes because we only used reading here. Reading a nil map is allowed.

pkg/controller/virtualmachine/virtualmachine.go
```diff
@@ -0,0 +1,475 @@
...
+		if err != nil {
+			return fmt.Errorf("error listing vgpudevices from cache: %v", err)
+		}
+		vGPUMap = buildVGPUMap(deviceList)
+	}
+	// map to hold device details
+	hostDeviceMap := make(map[string][]string)
+	gpuMap := make(map[string][]string)
+	for _, device := range vmi.Spec.Domain.Devices.HostDevices {
+		val, ok := envMap[util.ResourceNameToEnvVar(deviceplugins.PCIResourcePrefix, device.DeviceName)]
```
Yu-Jack, line 148: Just confirm that. Since the usbdevice is not like pcidevice and vGPU device are randomly picked up by KubeVirt, so we can skip collecting envMap from `USBResourcePrefix`. If yes, do we need to add some comments about why we don't need to collect from `USBResourcePrefix`?

pkg/controller/virtualmachine/virtualmachine.go
```diff
@@ -0,0 +1,475 @@
...
+	}
+	if vmObj.Annotations == nil {
+		return nil
+	}
+
+	_, ok := vmObj.Annotations[v1beta1.DeviceAllocationKey]
+	// no key, nothing is needed
+	if !ok {
+		return nil
+	}
```
Yu-Jack, line 261: NIT: Reading a nil map is allowed, we could remove it because we don't update the map later in this function.

Yu-Jack (approved):
LGTM, and just a reminder. We need to raise a chart PR for updating the service account permission to allow that pcidevices-controller can list/update virtualmachines and list virtualmachineinstances.

pkg/controller/virtualmachine/virtualmachine.go
ibrokethecloud: yes for usb devices a new plugin is created for each device so it will not suffer from this issue

pkg/controller/virtualmachine/virtualmachine.go
```diff
@@ -0,0 +1,462 @@
...
+
+// if multiple devices of same type are added to a VM, there may be duplicates in the deviceDetails values
+// since env variable would have been looked up twice from envMap and added to deviceDetails
+// so we dedup the unique device names
+func dedupDevices(deviceMap map[string][]string) map[string][]string {
+	for key, val := range deviceMap {
+		deviceMap[key] = slices.Compact(val)
+	}
+	return deviceMap
+}
```
WebberHuang1118, line 304: Jack and I found the duplicate in the annotation, it seems not expected?
image_720

WebberHuang1118 (commented):
LGTM, just a nit, thanks.

pkg/controller/virtualmachine/virtualmachine.go
Yu-Jack: FYI, it happened when we attached two vGPU devices to a VM.

ibrokethecloud: i can fix it up

ibrokethecloud: i have identified the reason why the names were being duplicated. I have now fixed this up and the additional commit contains the fix. I have kept the commits separate different to make it easier to track the changes. I will squash them once we are happy with the new set of changes.

WebberHuang1118 (approved):
LGTM, thanks for the PR

PR #94, opened 2024-10-08
feat: remove orphaned USB device automatically and add reconcile after deleting claim

Yu-Jack:

**Problem:**
Steps:

- Plugin USB device
- Enable device (Don't need to attach it to a VM)
- Unplug USB device
- Plugin USB device again

There are two USB device listed on the page, one of them is orphaned.

**Solution:**
Case A: If that orphaned USB device is enabled and unattached, just remove it directly.
Case B: if that orphaned USB device is enabled and attached to a VM, just update status. I think GUI can show message to users in the future.

Additional fix is that deleting attached USB device claim should be forbidden.

**Related Issue:**
https://github.com/harvester/harvester/issues/6720

**Test plan:**
Follow above steps to test it.

### DIFF
--- a/pkg/controller/usbdevice/usbdevice_controller.go
+++ b/pkg/controller/usbdevice/usbdevice_controller.go
@@ -26,6 +26,8 @@ type DevHandler struct {
 	usbClaimClient ctldevicerv1vbeta1.USBDeviceClaimClient
 	usbCache       ctldevicerv1vbeta1.USBDeviceCache
 	usbClaimCache  ctldevicerv1vbeta1.USBDeviceClaimCache
+
+	reconcileSignal chan struct{}
 }
 
 type UsageList struct {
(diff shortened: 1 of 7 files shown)


ibrokethecloud (approved):
lgtm. thanks.

pkg/controller/usbdevice/usbdevice_controller.go
```diff
@@ -174,14 +180,32 @@ func (h *DevHandler) handleList(createList []*v1beta1.USBDevice, updateList []*v
...
 		}
 	}
 
-	for _, usbDevice := range deleteList {
+	for _, usbDevice := range usageList.orphanedList {
+		err := h.usbClaimClient.Delete(usbDevice.Name, &metav1.DeleteOptions{})
+
+		if err == nil {
+			continue
+		}
```
WebberHuang1118, line 195: Do we need to verify if the error is from webhook?

Yu-Jack: I don't think we need to do that. Regardless of any errors, we just need to mark it as orphaned since the USB device is already not present on the host.

WebberHuang1118 (approved):
LGTM, thanks for the PR.

PR #97, opened 2024-11-15
fix: check group map before enable pci device passthrough

WebberHuang1118:
**Problem:**
PCIDevice's device plugin uses an out-of-date group after node rebooting

**Solution:**
check the group map before enabling PCIDevice device plugin

**Related Issue:**
https://github.com/harvester/harvester/issues/6892

**Test plan:**
- Enabling a PCIDevice
- Creating a VM attach the PCIDevice
- After starting the VM successfully, rebooting the node
- Checking if the VM can start successfully across node rebooting

### DIFF
--- a/pkg/controller/pcideviceclaim/pcideviceclaim_controller.go
+++ b/pkg/controller/pcideviceclaim/pcideviceclaim_controller.go
@@ -24,6 +24,7 @@ import (
 	"github.com/harvester/pcidevices/pkg/config"
 	"github.com/harvester/pcidevices/pkg/deviceplugins"
 	v1beta1gen "github.com/harvester/pcidevices/pkg/generated/controllers/devices.harvesterhci.io/v1beta1"
+	"github.com/harvester/pcidevices/pkg/iommu"
 )
 
 var (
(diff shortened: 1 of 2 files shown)


ihcsim:
For my own understanding, if the device's `status.iommuGroup` is stale, should it be updated to match that found in the group path, like what the pci device controller is doing in https://github.com/harvester/pcidevices/blob/c65a0933fd358d7fb631bc62847a5041418ff8f0/pkg/controller/pcidevice/pcidevice_controller.go#L98

Looks like the pci device controller is also doing the same group map check in https://github.com/harvester/pcidevices/blob/c65a0933fd358d7fb631bc62847a5041418ff8f0/pkg/controller/pcidevice/pcidevice_controller.go#L47-L52.

ibrokethecloud:
@ihcsim the issue is when the iommu group is stale, likely after an OS upgrade, the device plugin gets setup with the wrong iommu group, which results in wrong iommu group being passed to the launcher pod.

With this change we wait for the periodic device reconcile to update the pcidevice object and this blocks an incorrect device plugin from being setup.

ibrokethecloud (approved):
lgtm. thanks.

Yu-Jack (approved):

PR #100, opened 2024-12-16
feat: check usb/pci/gpu resource name

Yu-Jack:

**Problem:**
When creating or updating a non-existed pci/usb resource name, it's allowed. Besides, vGPU device is not checked either.

**Solution:**
We should check `.DeviceName` in VM spec while creating or updating. BTW, vGPU device is in the other field, so we need to check it in `spec.template.domain.devices.gpus`.

> .DeviceName means **resource name** in k8s.

**Related Issue:**
https://github.com/harvester/harvester/issues/6542

**Test plan:**

#### Test PCI Device

1. Enable pcidevices-controller Add-on
2. Enable one of PCI devices
3. Create a VM and attach previous PCI device
4. Click "Edit as YAML" button to manually edit YAML file, then click "Create" button.
5. Modify the `.DeviceName` in the spec.template.domain.devices.hostDevices.

If you're not sure how to do it, please refer following demo recording.

https://github.com/user-attachments/assets/fd2336e6-8bc9-49ba-81e9-a23661ede5a5

#### Test USB Device

1. Enable pcidevices-controller Add-on
2. Enable one of USB devices
3. Create a VM and attach previous USB device
4. Click "Edit as YAML" button to manually edit YAML file, then click "Create" button.
5. Modify the `.DeviceName` in the spec.template.domain.devices.hostDevices.

If you're not sure how to do it, please refer following demo recording.

https://github.com/user-attachments/assets/3f548d07-d4e0-48c2-a5be-5287b99eadcc

### DIFF
--- a/pkg/webhook/vm_validatation.go
+++ b/pkg/webhook/vm_validatation.go
@@ -15,8 +15,9 @@ import (
 type vmDeviceHostValidator struct {
 	types.DefaultValidator
 
-	usbCache v1beta1.USBDeviceCache
-	pciCache v1beta1.PCIDeviceCache
+	usbCache  v1beta1.USBDeviceCache
+	pciCache  v1beta1.PCIDeviceCache
+	vgpuCache v1beta1.VGPUDeviceCache
 }
 
 func (vmValidator *vmDeviceHostValidator) Resource() types.Resource {
@@ -33,10 +34,11 @@ func (vmValidator *vmDeviceHostValidator) Resource() types.Resource {
 	}
 }
 
-func NewDeviceHostValidation(usbCache v1beta1.USBDeviceCache, pciCache v1beta1.PCIDeviceCache) types.Validator {
+func NewDeviceHostValidation(usbCache v1beta1.USBDeviceCache, pciCache v1beta1.PCIDeviceCache, vgpuCache v1beta1.VGPUDeviceCache) types.Validator {
 	return &vmDeviceHostValidator{
-		usbCache: usbCache,
-		pciCache: pciCache,
+		usbCache:  usbCache,
+		pciCache:  pciCache,
+		vgpuCache: vgpuCache,
 	}
 }
 
@@ -172,9 +174,22 @@ func (vmValidator *vmDeviceHostValidator) validateUSBDevice(resourceName string)
 	return true, nil
 }
 
+func (vmValidator *vmDeviceHostValidator) validateVGPUDevice(resourceName string) (found bool, err error) {
+	vGPUDeviceObjs, err := vmValidator.vgpuCache.GetByIndex(vGPUDeviceByResourceName, resourceName)
+	if err != nil {
+		return false, fmt.Errorf("error looking up vGPU device %s from cache: %v", resourceName, err)
+	}
+
+	if len(vGPUDeviceObjs) == 0 {
+		return false, nil
+	}
+
+	return true, nil
+}
+
 func (vmValidator *vmDeviceHostValidator) validateGPUs(vmObj *kubevirtv1.VirtualMachine) error {
 	for _, gpu := range vmObj.Spec.Template.Spec.Domain.Devices.GPUs {
-		foundInPCI, err := vmValidator.validatePCIDevice(gpu.DeviceName)
+		foundInPCI, err := vmValidator.validateVGPUDevice(gpu.DeviceName)
 
 		if err != nil {
 			return err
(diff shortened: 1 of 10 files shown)


pkg/webhook/vm_validatation.go
```diff
@@ -108,3 +123,67 @@ func (vmValidator *vmDeviceHostValidator) validateDevicesFromSameNodes(vmObj *ku
...
+		}
+
+		if !foundInPCI && !foundInUSB {
+			return fmt.Errorf("hostdevice %s: resource name %s not found in pcidevice and usbdevice cache", hostDevice.Name, hostDevice.DeviceName)
+		}
+	}
+	return nil
+}
+
+func (vmValidator *vmDeviceHostValidator) validatePCIDevice(resourceName string) (found bool, err error) {
```
bk201, line 149: these validators check if there is a pci resource that exists in the cluster. Do we need to check the resource with a node in any situation?

Yu-Jack: I think we need to check below situations.

1. The node selector is different from pci/usb device's node
2. The resource is used by other VMs

Currently, we don't have those checks in backend code.

pkg/webhook/vm_validatation.go
```diff
@@ -108,3 +123,67 @@ func (vmValidator *vmDeviceHostValidator) validateDevicesFromSameNodes(vmObj *ku
...
+	for _, gpu := range vmObj.Spec.Template.Spec.Domain.Devices.GPUs {
+		foundInPCI, err := vmValidator.validatePCIDevice(gpu.DeviceName)
+
+		if err != nil {
+			return err
+		}
+
+		if !foundInPCI {
+			return fmt.Errorf("gpu device %s: resource name %s not found in pcidevice cache", gpu.Name, gpu.DeviceName)
+		}
```
WebberHuang1118, line 185: Hi Jack,
Why do we validate GPUs with PCI devices?

Yu-Jack: As I know, the GPU is also bound to PCI device, so we can use PCI device to check it.

Yu-Jack: I thought we just mapped the vGPU to PCI device. It's like USB device. We can pass whole USB controller to VM with PCI device, or just pass one of USB ports to VM with USB passthrough.

pkg/webhook/vm_validatation.go
Yu-Jack: I think we could keep this issue simple. So, I'll only check if the resource name exists in the cluster. For your case, if we decide to check node selector, we should consider node affinity. Besides, it's valid that same resource names are used by different VMs. We also need to consider checking if the resource is enough.

pkg/webhook/vm_validatation.go
```diff
@@ -108,3 +125,80 @@ func (vmValidator *vmDeviceHostValidator) validateDevicesFromSameNodes(vmObj *ku
 
 	return nil
 }
+
+func (vmValidator *vmDeviceHostValidator) validateHostDevices(vmObj *kubevirtv1.VirtualMachine) error {
+	for _, hostDevice := range vmObj.Spec.Template.Spec.Domain.Devices.HostDevices {
```
WebberHuang1118, line 130: The object access path `vmObj.Spec.Template.Spec.Domain.Devices.HostDevices` is quite long, maybe we can consider extracting it as a utility function.

pkg/webhook/vm_validatation.go
```diff
@@ -108,3 +125,80 @@ func (vmValidator *vmDeviceHostValidator) validateDevicesFromSameNodes(vmObj *ku
...
+
+	if len(vGPUDeviceObjs) == 0 {
+		return false, nil
+	}
+
+	return true, nil
+}
+
+func (vmValidator *vmDeviceHostValidator) validateGPUs(vmObj *kubevirtv1.VirtualMachine) error {
+	for _, gpu := range vmObj.Spec.Template.Spec.Domain.Devices.GPUs {
```
WebberHuang1118, line 191: ditto

WebberHuang1118 (approved):
LGTM, just a nit, thanks.

pkg/webhook/vm_validatation.go
Yu-Jack: We can extract it later if we truly utilize a significant number of these paths.

ibrokethecloud (approved):
lgtm. thanks.

bk201 (approved):

Yu-Jack:
@Mergify backport v1.3

PR #107, opened 2025-03-12
minor fix for handling vGPUs when using NVIDIA RTX5000 Ada devices

ibrokethecloud:

**Problem:**

Some nvidia vGPU profiles have a lower case character in their name.
This breaks vgpu device passthrough to the workloads as the kubevirt cr permitted devices does not match the actual vgpu profile.

**Solution:**

minor change to vgpu device name generation where we do not force conversion to upper case characters as some devices like RTX5000 ADA preent the vgpu types as RTX5000-Ada which results in device plugin name not matching the actual gpu profile

**Related Issue:**
https://github.com/harvester/harvester/issues/6294
**Test plan:**


Yu-Jack (approved):
Thanks!

WebberHuang1118 (approved):
LGTM, thanks.

ibrokethecloud:
@mergifyio backport v1.5

PR #123, opened 2025-07-27
k8s + wrangler version bump

ibrokethecloud:

**Problem:**

PR bumps k8s and wrangler related dependencies

In addition, there are minor changes to reconcile logic and test cases as a client.Get call no longer returns a `nil` object when object is not found but an `empty` object which results in issues where the code base is dependent on object being `nil` to make certain decisions

**Solution:**

**Related Issue:**

**Test plan:**


pkg/controller/gpudevice/gpu_controller.go
```diff
@@ -132,18 +132,29 @@ func (h *Handler) SetupSRIOVGPUDevices() error {
...
 		}
+
+		//wrangler bump to v3.1.
 		if existingClaim != nil {
 			// pciDeviceClaim exists skipping
 			logrus.Debugf("skipping creation of vGPUDevice %s as PCIDeviceClaim exists", existingClaim.Name)
 			continue
 		}
 
+		fmt.Printf("creating device: %s\n", v.Name)
```
Vicente-Cheng, line 157: Redundant? Or moving to logger?

Vicente-Cheng (approved):
just a nit, others LGTM

pkg/controller/gpudevice/gpu_controller.go
ibrokethecloud: redundant

Vicente-Cheng (approved):
thanks

Vicente-Cheng:
@mergifyio backport v1.6

PR #128, opened 2025-08-25
Fix bug in vmi deletion

liujiong63:
**Problem:**
The controller keeps re-queuing vmi CR when deleting vm CR

**Solution:**
Do not re-queuing vmi CR when vm CR is not found

**Related Issue:**
https://github.com/harvester/harvester/issues/8954

**Test plan:**
1. Create a pcideviceclaim CR
2. Create a vm and configure hostDevice associated with the pcideviceclaim, start the vm
3. Delete the vm, controller should not re-queuing vmi CR


pkg/controller/virtualmachine/virtualmachine.go
```diff
@@ -364,7 +365,12 @@ func (h *Handler) OnVMIDeletion(_ string, vmi *kubevirtv1.VirtualMachineInstance
 
 	vmObj, err := h.vmCache.Get(vmi.Namespace, vmi.Name)
 	if err != nil {
-		return vmi, fmt.Errorf("error fetching vm object for vmi %s/%s: %v", vmi.Namespace, vmi.Name, err)
+		if k8serrors.IsNotFound(err) {
+			// vm is deleted
```
w13915984028, line 369: @liujiong63 Thanks for the fix.

In the `delete VM case`, could you add one more check on L375, which detects the `vmObj.DeletionTimestamp != nil` and return nil.

w13915984028 (approved):
LGTM, thanks.

pkg/controller/virtualmachine/virtualmachine.go
```diff
@@ -364,7 +365,15 @@ func (h *Handler) OnVMIDeletion(_ string, vmi *kubevirtv1.VirtualMachineInstance
...
+			return vmi, nil
+		} else {
+			return vmi, fmt.Errorf("error fetching vm object for vmi %s/%s: %v", vmi.Namespace, vmi.Name, err)
+		}
```
Yu-Jack, line 373:
```suggestion
		if k8serrors.IsNotFound(err) {
			// vm is deleted
			return vmi, nil
		}
		return vmi, fmt.Errorf("error fetching vm object for vmi %s/%s: %v", vmi.Namespace, vmi.Name, err)
```

Yu-Jack: In order to pass golint.

Yu-Jack (changes requested):

Yu-Jack (approved):
LGTM

ibrokethecloud (approved):
lgtm. thanks.

PR #174, opened 2026-03-31
fix parent sriovgpudevice label on PCIdevice object

ibrokethecloud:

**Problem:**

PR https://github.com/harvester/pcidevices/pull/165 added additional logic in vGPUdevice setup, to add the parent SRIOVGPUDevice name as a label to the PCIDevice. This is needed to ensure that the UI can easily filter PCIDevices which are owned by a specific SRIOVGPUDevice.

The label value being added was wrong, and was the vGPUDevice name when it needed to be the parent SRIOVGPUDevice.

**Solution:**

The PR fixes the correct parent SRIOVGPUDevice label value and also enables golint-ci installation in dapper builds.

**Related Issue:**
https://github.com/harvester/harvester/issues/10120#issuecomment-4159206317

**Test plan:**

### DIFF
--- a/Dockerfile.dapper
+++ b/Dockerfile.dapper
@@ -29,7 +29,7 @@ RUN export K8S_VERSION=v1.34.0 && \
     tar -C /usr/local/kubebuilder --strip-components=1 -zvxf envtest-bins.tar.gz
 
 ## install golangci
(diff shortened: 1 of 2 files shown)


WebberHuang1118 (approved):
LGTM, thanks.

ibrokethecloud:
@mergify backport v1.8

PR #179, opened 2026-04-15
fix vgpu status reconcile after reboots and upgrades

ibrokethecloud:

**Problem:**

vgpu device cannot be enabled after upgrade as the vgpu type is not valid as per status
```
hp-130-tink-system-000008006': handler on-vgpu-change: VGPUType specified NVIDIA A2-2A is not available for vGPU 0000:08:00.6 
```

**Solution:**

change logic for vgpu status update in reconcileVGPUSetup, to allow s…ysfs status to be updated to k8s object to ensure device gets re-enabled after upgrades and reboots

**Related Issue:**
https://github.com/harvester/harvester/issues/10394#top

**Test plan:**


Yu-Jack (approved):
LGTM

ibrokethecloud:
@mergify backport v1.8

PR #182, opened 2026-04-29
feat: add identifying usb device class

Yu-Jack:

**Problem:**
In current USB device page, we don't show the USB type on it, which users might not know what it's.

image

**Solution:**
Use https://www.usb.org/defined-class-codes to identify the USB type.

If bDeviceClass is `00`
-> Look up the value of bInterfaceClass and map it to the table from the above reference.
if bDeviceClass is not `00`
-> Directly look up the value of bDeviceClass and map it to the table from the above reference.

Besides, read `product` file and append the information after the description.

```
# Example for storage device
apiVersion: devices.harvesterhci.io/v1beta1
kind: USBDevice
metadata:
  creationTimestamp: "2026-04-29T04:34:10Z"
  generation: 1
  labels:
    nodename: harvester1
  name: harvester1-0951-1666-002045
  resourceVersion: "5515884"
  uid: e1e2cf8e-b305-43df-b832-7dc48ace7822
status:
  classType: Mass Storage
  description: DataTraveler 100 G3/G4/SE9 G2/50 Kyson (Kingston Technology) [DataTraveler
    3.0]
  devicePath: /dev/bus/usb/002/045
  enabled: false
  nodeName: harvester1
  pciAddress: "0000:02:01.0"
  productID: "1666"
  resourceName: kubevirt.io/harvester1-0951-1666-002045
  vendorID: "0951"
```

**Related Issue:**
Issue: https://github.com/harvester/harvester/issues/8330
UI: https://github.com/harvester/harvester-ui-extension/pull/839

**Test plan:**

It also works to existed USB devices.

image

### DIFF
--- a/Dockerfile.dapper
+++ b/Dockerfile.dapper
@@ -34,7 +34,7 @@ RUN go install github.com/golangci/golangci-lint/v2/cmd/golangci-lint@v2.11.4
 ENV GO111MODULE=on
 ENV DAPPER_ENV="REPO TAG DRONE_TAG"
 ENV DAPPER_SOURCE="/go/src/github.com/harvester/pcidevices/"
-ENV DAPPER_OUTPUT="./bin ./dist ./pkg"
+ENV DAPPER_OUTPUT="./bin ./dist ./pkg ./charts"
 ENV DAPPER_RUN_ARGS="-v /proc:/host/proc --privileged"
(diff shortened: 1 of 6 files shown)


ibrokethecloud (approved):
lgtm. thanks for the enhancement.

PR #186, opened 2026-05-07
fix: update existed usb device in upgrade case

Yu-Jack:

**Problem:**
For those exited usb devices, their description and class type aren't changed in upgrade case.

**Solution:**
if description and class type is changed, update it.

**Related Issue:**
https://github.com/harvester/harvester/issues/8330

**Test plan:**

Before:

https://github.com/user-attachments/assets/8a302d0e-3944-4686-882b-59155e9c4a46

After:

https://github.com/user-attachments/assets/5669ec1b-fce8-4310-b35b-7a3027ece1f5


Yu-Jack:
I also tested this case, and It works well:

1. enable old pcidevices-controller
2. enable and passthrough usb device to vm
3. upgrade pcidevices-controller image to this fix version
4. check whether vm can read the data from the usb device

ibrokethecloud (approved):

PR #189, opened 2026-05-20
Fix: Trim trailing separators from generated PCI resource names

skanakal:

**Problem:**

The [resourceName()](https://github.com/harvester/pcidevices/blob/master/pkg/apis/devices.harvesterhci.io/v1beta1/pcidevice.go#L94) helper could generate invalid Kubernetes extended resource names when PCI product names from the pci.ids database contain trailing spaces inside the bracket notation, which voilates the k8s extended resource naming rules because resource name must end with an alphanumeric character...

**Solution:**

It trims the trailing separators 

**Related Issue:**
https://github.com/harvester/harvester/issues/10108

**Test plan:**

unit tests added.


skanakal:
@ibrokethecloud Could you please review this PR when you have time?

ibrokethecloud (approved):

Yu-Jack (approved):
Thanks for the PR.

PR #198, opened 2026-06-19
PR drops dapper and moves to docker build stage builds for running ci

ibrokethecloud:

**Problem:**

PR drops dapper in favor of using docker multi stage builds

**Solution:**

**Related Issue:**
https://github.com/harvester/harvester/issues/10617

**Test plan:**


ibrokethecloud:
2nd commit adds a .hadolint.yaml which inherits the codefactor base hadolint configuration file https://github.com/codefactor-io/default-configs/blob/master/.hadolint.yaml and adds https://github.com/hadolint/hadolint/wiki/DL3037 to the ignored list

bk201 (approved):

.hadolint.yaml
```diff
@@ -0,0 +1,235 @@
+ignored:
```
Yu-Jack, line 1: Just curious, why do we need this? and why is it so many?

Yu-Jack (approved):
LGTM

.hadolint.yaml
ibrokethecloud: i copied the `hadolint.yaml` from codefactor and appended the extra version pinning one. else codefactor would fail and insist on having pinned versions of all packages used in dockerfile. 

https://www.codefactor.io/repository/github/harvester/pcidevices/pull/fixed/198

PR #260, opened 2026-06-26
changes needed to support container workloads

ibrokethecloud:

**Problem:**

Support baremetal gpu workloads on existing Harvester clusters.

The KVM/AIE drivers cannot be used with baremetal workloads as a result we need a means to allow users to run vGPU / baremetal workloads on the same cluster.

**Solution:**

The PR adds changes to pcidevices controller to allow users where baremetal gpu workloads need to be run with label `harvesterhci.io/gpu-baremetal-workloads=true`. 

The pcidevice controller skips creation of sriovgpudevices and vgpudevices on nodes containing this label. On existing nodes with sriovgpudevices and vgpudevices, the existing devices are removed.

The new rule is also used in nvidia runtime container chart https://github.com/harvester/charts/pull/563 to skip nodes containing the additional label. 

With these changes in place users can now deploy nvidia gpu operator on their existing Harvester clusters to leverage gpus with baremetal container workloads.

**Related Issue:**
https://github.com/harvester/harvester/issues/5321

**Test plan:**


pkg/webhook/node.go
```diff
@@ -0,0 +1,81 @@
...
+	}
+
+	if newNodeObj.Labels[v1beta1.GPUContainerWorkloadKey] == v1beta1.GPUContainerWorkloadValue {
+		newDisableGPU = true
+	}
+
+	// if label has been added on node, then we need to verify there are no enabled sriovGPUDevices on this node
+	if !oldDisableGPU && newDisableGPU {
+		return v.CheckEnabledSRIOVGPUDevices(newNodeObj.Name)
+	}
```
Yu-Jack: Do we need to check the opposite direction?

For example, users might have requested gpu for their contains, and they accidentally delete the workload label?

Yu-Jack (commented):
Overall LGTM. Thanks! Just have a question.

pkg/webhook/node.go
ibrokethecloud: the node label can only be managed by cluster admins. you will also need to add extra gpu-operator labels to skip node from being managed by the gpu operator https://github.com/harvester/harvester/pull/11046

Yu-Jack: I mean:

let's say we have one node (A)

- when uses label `harvesterhci.io/gpu-baremetal-workloads: "true"` on node A
- user starts using gpu with container on node A

Should we block the request that deletes `harvesterhci.io/gpu-baremetal-workloads: "true"` label on the webhook because some containers are using gpu?

ibrokethecloud: I can add a check to block pods with `nvidia.com/gpu` resource usage on the node.

```
apiVersion: v1
kind: Pod
metadata:
  name: nbody-gpu-benchmark
  namespace: default
spec:
  restartPolicy: OnFailure
  # runtimeClassName: nvidia <== Only needed for v25.3.x
  containers:
  - name: cuda-container
    image: nvcr.io/nvidia/k8s/cuda-sample:nbody
    args: ["nbody", "-gpu", "-benchmark"]
    resources:
      limits:
        nvidia.com/gpu: 1
```

I am not sure yet though if the gpu operator allows custom name to be used when advertising the devices on a node.

pkg/controller/nodes/node_controller.go
```diff
@@ -100,6 +102,13 @@ func (h *handler) reconcileNodeDevices(name string, node *v1beta1.Node) (*v1beta
 		return node, nil
 	}
 
+	var disableGPU bool
```
WebberHuang1118: The naming `disableGPU` is a bit confusing to me. IIUC, `baremetalGPU` might be more appropriate. Suggest changing this naming throughout this PR.

WebberHuang1118 (dismissed):
LGTM in general.

A suggestion about the naming, and I have the same question as @Yu-Jack in [this comment](https://github.com/harvester/pcidevices/pull/260/changes#r3503248460), thanks.

pkg/webhook/node.go
Yu-Jack: I saw this. I guess they may change it over time on demand.

```yaml
#Config map 
  version: v1
  sharing:
    timeSlicing:
      renameByDefault: true  
      resources:
        - name: nvidia.com/gpu
          replicas: 4
```


``` yaml
# Pod:
 resources:
    limits:
      nvidia.com/gpu.shared: 1
```

reference: https://docs.nvidia.com/datacenter/cloud-native/gpu-operator/latest/gpu-sharing.html#about-configuring-gpu-time-slicing

Yu-Jack: Maybe we can use strings.Contains("nvidia.com/gpu") to check.

The only concern is that I hope they won't release like `nvidia.com/multi-gpu` LOL

ibrokethecloud: The latest commit adds the extra logic to identify running pods consuming the gpu device. the device name can be overridden via an env variable. I will submit a chart PR to allow exposing this via the helm chart values, which will allow users to customize pcidevices controller addon via the addon yaml manifest

WebberHuang1118 (dismissed):

pkg/controller/gpudevice/gpu_controller.go
```diff
@@ -133,16 +133,20 @@ func (h *Handler) OnGPUChange(_ string, gpu *v1beta1.SRIOVGPUDevice) (*v1beta1.S
 }
 
 // SetupSRIOVGPUDevices is called by the node controller to reconcile objects on startup and predefined intervals
-func (h *Handler) SetupSRIOVGPUDevices() error {
+func (h *Handler) SetupSRIOVGPUDevices(disableGPU bool) error {
```
Yu-Jack: Do we also need to change `disableGPU` to `baremetalGPU` here?

Yu-Jack (approved):
Thanks.

pkg/controller/gpudevice/gpu_controller.go
ibrokethecloud: this has been renamed

ibrokethecloud:
@mergify backport v1.9

PR #279, opened 2026-07-13
Add CODEOWNERS file

Vicente-Cheng:
#### Problem:
Add CODEOWNER

#### Solution:
Add CODEOWNER

#### Related Issue(s):
https://github.com/harvester/harvester/issues/9607


Vicente-Cheng:
Re-opening from fork instead of upstream branch.

PR #296, opened 2026-07-22
track nvidia gpus to block pcidevice claims when node is used for baremetal gpu workloads

ibrokethecloud:
**IMPORTANT: Please do not create a Pull Request without creating an issue first.**

**Problem:**

Based on feedback for baremetal gpu container workload HEP review it was suggested that pcideviceclaims for nvidia gpus on nodes being used for container workloads should be blocked.

**Solution:**

the node controller now reconciles nvidia gpu device address in the node status.

changes to pcidevices controller to track nvidia gpu's in node.devices object, to allow webhook to block pcideviceclaim creation for nvidia gpus if the node is labelled for baremetal gpu container workloads

**Related Issue:**
https://github.com/harvester/harvester/issues/5321

**Test plan:**

PR #313, opened 2026-08-07
change node.Status.GPUAddresses field to mandatory to ensure successful reconcile

ibrokethecloud:

**Problem:**

PCIDevices reconcile is broken on new install due to the PR https://github.com/harvester/pcidevices/pull/296

PR changes the `node.devices` object to make `GPUAddresses` a mandatory field in `status`.

Currently when `[]GPUAddresses` are empty, since field is optional, nothing is set in status field of object.

DeepEqual always finds `nil` and `[]string` do not match so keeps triggering an update which results in node reconcile never moving on to subsequent device reconcile.

**Solution:**

**Related Issue:**
https://github.com/harvester/harvester/issues/11360

**Test plan:**

* Install harvester master-head or v1.9-head to a new cluster
* Enable PCIDevices controller on a node without GPUs
* Wait for a few mins, and PCIDevices should be populated


Vicente-Cheng (approved):
thanks for the fix!

ibrokethecloud:
@mergify backport v1.9

